PR #35, opened 2024-08-26
Add policies page

### LINKED ISSUE #17: Policy listing
Support for listing Kuadrant policies (`AuthPolicy`, `DNSPolicy`, `TLSPolicy`, `RateLimitPolicy`) via the UI. 

The listing view is currently in the POC but we need to update with new UI mocks.

General list all policies view:
* https://www.figma.com/proto/r6rh1Jkx9vI9G2pXv0rT2D/OCP-Dynamic-plugin-(GA)?node-id=11-2203&starting-point-node-id=12%3A4426&show-proto-sidebar=1

- In the `Developer` perspective, users should be able to list: `AuthPolicy`, `RateLimitPolicy`
- In the `Administrator` perspective, users should be able to list all types of policy
- Project/NS selector should be included in these views

david-martin:
Closes #17 

TODO:

- [x] add policy type tab selector to top of page, including 'all policies' tab
- [x] implement single policy type view
- [x] implement all policies view
- [x] add name/label filter input to top of page
- [x] add kebab menu for each policy item in list views - link to edit/delete etc... (initially no link if pages don't exist) (based on example at https://www.patternfly.org/components/menus/dropdown#with-kebab-toggle)
- [x] add blue info box to top of page
- [x] review fields being show for each policy type in list view, ensuring most relevant/useful are shown - Using 'Created' as a formatted timestamp (like other resource lists in console) and a Status conditions summary string e.g. "Accepted=True,Enforced=True"
- [x] fix 404 when a single project is selected in project dropdown (works for 'all projects') - No longer 404s, but always routes back to the 'All policies' tab. Deferring how to get it to stick to the open tab when switching project context for now.
- [x] copy or reuse page for both admin & dev perspective, with just auth/ratelimit shown for dev perspective
- [ ] ~fix page reload issue when typing in filter input~ (defer at this time, low priority)

### DIFF
--- a/src/components/KuadrantPoliciesPage.tsx
+++ b/src/components/KuadrantPoliciesPage.tsx
@@ -1,0 +1,423 @@
+import * as React from 'react';
+import { useParams } from 'react-router-dom';
+import { useTranslation } from 'react-i18next';
+import { Dropdown, DropdownItem, DropdownList, MenuToggle, MenuToggleElement, Alert, AlertGroup, Title} from '@patternfly/react-core';
+import { sortable } from '@patternfly/react-table';
+import EllipsisVIcon from '@patternfly/react-icons/dist/esm/icons/ellipsis-v-icon';
+import { useK8sWatchResource, K8sResourceCommon, ResourceLink, useActiveNamespace, HorizontalNav, useListPageFilter,
+  ListPageBody,
+  ListPageCreate,
+  ListPageFilter,
+  VirtualizedTable,
+  TableData,
+  RowProps,
+  TableColumn,
+  NamespaceBar,
+  Timestamp,
+  useActivePerspective,
+ } from '@openshift-console/dynamic-plugin-sdk';
+import './kuadrant.css';
+
+interface Resource {
+  name: string;
+  gvk: {
+    group: string;
+    version: string;
+    kind: string;
+  };
+}
+
+interface ExtendedK8sResourceCommon extends K8sResourceCommon {
+  status?: {
+    addresses?: { value: string }[];
+  };
+}
+
+const statusConditionsAsString = (obj: any) => {
+  if (!obj.status || !obj.status.conditions) {
+    return '';
+  }
+  return obj.status.conditions
+    .map(condition => `${condition.type}=${condition.status}`)
+    .join(',');
+}
+
+const resources: Resource[] = [
+  { name: 'AuthPolicies', gvk: { group: 'kuadrant.io', version: 'v1beta2', kind: 'AuthPolicy' } },
+  { name: 'DNSPolicies', gvk: { group: 'kuadrant.io', version: 'v1alpha1', kind: 'DNSPolicy' } },
+  { name: 'RateLimitPolicies', gvk: { group: 'kuadrant.io', version: 'v1beta2', kind: 'RateLimitPolicy' } },
+  { name: 'TLSPolicies', gvk: { group: 'kuadrant.io', version: 'v1alpha1', kind: 'TLSPolicy' } },
+];
+
+type AllPoliciesTableProps = {
+  data: K8sResourceCommon[];
+  unfilteredData: K8sResourceCommon[];
+  loaded: boolean;
+  loadError: any;
+};
+
+const AllPoliciesTable: React.FC<AllPoliciesTableProps> = ({ data, unfilteredData, loaded, loadError }) => {
+  const { t } = useTranslation();
+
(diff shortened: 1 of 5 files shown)


david-martin:
@jasonmadigan Rebased and ready for review.

jasonmadigan:
👀

src/components/KuadrantPoliciesPage.tsx
```diff
@@ -0,0 +1,422 @@
...
+  };
+}
+
+interface ExtendedK8sResourceCommon extends K8sResourceCommon {
+  status?: {
+    addresses?: { value: string }[];
+  };
+}
+
+const statusConditionsAsString = (obj: any) => {
```
jasonmadigan, line 36: not urgent to move or genericise, but this might be useful elsewhere. mentally noting so I know where to nab.

src/components/KuadrantPoliciesPage.tsx
```diff
@@ -0,0 +1,422 @@
...
+  if (!obj.status || !obj.status.conditions) {
+    return '';
+  }
+  return obj.status.conditions
+    .map(condition => `${condition.type}=${condition.status}`)
+    .join(',');
+}
+
+const resources: Resource[] = [
+  { name: 'AuthPolicies', gvk: { group: 'kuadrant.io', version: 'v1beta2', kind: 'AuthPolicy' } },
```
jasonmadigan, line 46: wondering to myself how we should generally approach our CRD versioning and our UI. I guess the UI will support certain versions, or perhaps be in lock step. wondering what we do with resources at different versions and the UI. expect we won't have too many differences in terms of what we show in tables and the like, but worth thinking about.

src/components/KuadrantPoliciesPage.tsx
```diff
@@ -0,0 +1,422 @@
...
+const DropdownWithKebab: React.FC<DropdownWithKebabProps> = ({ obj }) => {
+  const [isOpen, setIsOpen] = React.useState(false);
+
+  const onToggleClick = () => {
+    setIsOpen(!isOpen);
+  };
+
+  const onSelect = (_event: React.MouseEvent<Element, MouseEvent> | undefined, value: string | number | undefined) => {
+    // eslint-disable-next-line no-console
+    console.log('selected', value);
```
jasonmadigan, line 145: must see what other plugins are doing with logging. Some of the core console stuff seems to be using some sort of fancy/decorated logger, perhaps with levels. just a thought, requesting no action

src/components/KuadrantPoliciesPage.tsx
```diff
@@ -0,0 +1,422 @@
...
+          isExpanded={isOpen}
+        >
+          <EllipsisVIcon />
+        </MenuToggle>
+      )}
+      shouldFocusToggleOnSelect
+    >
+      <DropdownList>
+        <DropdownItem value={0} key="edit">
+          Edit
```
jasonmadigan, line 169: no need right now, but we may want to look at using the built-in i8n stuff for strings (Example in the old dashboard view)

src/components/KuadrantPoliciesPage.tsx
```diff
@@ -0,0 +1,422 @@
...
+          </Alert>
+        </AlertGroup>
+        <div className='co-m-nav-title--row'>
+          <ListPageFilter
+            data={data}
+            loaded={loaded}
+            onFilterChange={onFilterChange}
+          />
+          <ListPageCreate groupVersionKind={resource.gvk}>{t(`plugin__console-plugin-template~Create ${resource.gvk.kind}`)}</ListPageCreate>
+        </div>        <PoliciesTable
```
jasonmadigan, line 330: pikcing nits: maybe newline

jasonmadigan (approved):
few some thoughts as I went through, but nothing that'd make me hesitate to merge

jasonmadigan:
@david-martin ^

src/components/KuadrantPoliciesPage.tsx
david-martin: Once we go v1, this will definitely be something to consider as soon as a new crd version is available.

I suspect the plugin code will keep in lockstep with the latest version (as long as that's what the 'stored' version is), and a conversion webhook would be required with the newer release of the operator to migrate older versions of crds to newer.
Both old & new versions would need to be supported for some time.

david-martin:
@jasonmadigan Thanks for review.
Ack on comments, with response on the CRD versioning one.

jasonmadigan:
LGTM

PR #47, opened 2024-09-06
Dnspolicy create

david-martin:
Part of #30 (the create part)

There's a lot in the create page because of the optional top level fields for healthCheck and loadBalancing (which is based on setting routingStrategy to loadbalanced).
The other top level fields are less complex:
- name
- namespace (uses a namespace dropdown)
- targetRef (uses a gateway dropdown by listing all gateways in the cluster)

TODO:

- [ ] namespace bar in top of page (which sets the namespace dropdown automatically on first load)
- [ ] alert modal on create error
- [ ] allow for matchLabels as an alternative to matchExpressions under  `weighted.custom[].selector`
- [ ] i18n in numerous places
- [ ] Radio options for Create via: form / yaml
- [ ] Link for how 'to create an additional Gateway'
- [ ] wire up 'Create DNSPolicy' button on policy listing page to this new page


src/components/gateway/GatewaySelect.tsx
```diff
@@ -0,0 +1,64 @@
...
+            key={index}
+            value={`${gateway.namespace}/${gateway.name}`}
+            label={`${gateway.namespace}/${gateway.name}`}
+            />
+          ))}
+        </FormSelect>
+        <FormHelperText>
+          <HelperText>
+            {/* TODO: link to 'here', where? */}
+            <HelperTextItem>Description here, to create an additional Gateway go to <a>here</a></HelperTextItem>
```
jasonmadigan, line 56: link to `/k8s/ns/ingress-gateway/gateway.networking.k8s.io~v1~Gateway/~new` I guess?

david-martin: yeah, seems reasonable.
I'll link to `/k8s/cluster/gateway.networking.k8s.io~v1~Gateway/~new`

src/components/KuadrantDNSPolicyCreatePage.tsx
```diff
@@ -0,0 +1,327 @@
...
+        healthCheck: healthCheck.endpoint ? healthCheck : undefined,
+      },
+    });
+  });
+
+  const history = useHistory();
+
+  React.useEffect(() => {
+    const updatedYamlResource = {
+      apiVersion: 'kuadrant.io/v1alpha1',
```
jasonmadigan, line 79: small thing, but I guess this applies to all of these views - should we have a helper to fetch the latest apiVersions for our resources? saves us having to deal with drift here (well, for the most part). I guess ditching the apiVersion would do the same.

src/components/KuadrantPoliciesPage.tsx
```diff
@@ -38,6 +37,7 @@ import {
   useActivePerspective,
 } from '@openshift-console/dynamic-plugin-sdk';
 import './kuadrant.css';
+import getModelFromResource from '../utils/getModelFromResource';
```
jasonmadigan, line 40: not a comment for this particular line, just that this isn't visible in the PR:


is this hooked up to the Create DNSPolicy button on the policies view?

jasonmadigan (approved):
LGTM.

PR #51, opened 2024-09-10
Generic create function

R-Lawton:
To use see example in https://github.com/Kuadrant/console-plugin/pull/52/files#diff-75017fdb2abfd575a647de98fdeac54a4156d480346c6a74f64c93542cdec41e


david-martin (approved):
Let's merge this and iterate as needed.
I'll try wire it up to dnspolicy create

PR #62, opened 2024-09-11
AuthPolicy create view

jasonmadigan:
Screenshot 2024-09-11 at 17 15 11


console-extensions.json
```diff
@@ -48,6 +48,13 @@
       "component": { "$codeRef": "PolicyTopologyPage" }
     }
   },
+  {
+     "type": "console.page/route",
+     "properties": {
+       "path": "/kuadrant/authpolicy/new",
```
david-martin, line 54: I'm not convinced we need custom routes for any that we 'override'... yet.
All links to create a new auth policy should render the component referenced further below

jasonmadigan: removed

Issue #59, opened 2024-09-11
Establish a pattern for reusing a policy create page for updates to a policy as well


david-martin:
Depending on what's landed on main, the DNSPolicy or TLSPolicy create pages could used as the reference to establish this pattern in.

Some thoughts on what would need to change in create views to allow reuse for updating:
- allow the initial resource to be populated in full on view load instead of having starting defaults only
- disallow name & namespace setting 
- change the create button at the bottom to 'Save' and 'Reload' buttons, similar to other views in the console
- call [k8sUpdateResource](https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/docs/api.md#k8supdateresource) instead

PR #70, opened 2024-09-16
Fix showing of server side validation error response in RLP & DNS cre…

### LINKED ISSUE #58: Implement and wire up a validation helper to validate policy fields before attempting to save the policy
Ideally part of this helper https://github.com/Kuadrant/console-plugin/blob/main/src/utils/createResource.tsx

If server side validation is relied on, the promise rejection from https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/docs/api.md#k8screateresource may be useable to do this.

david-martin:
…ate views

closes #58 

The Create button on the DNS & RLP pages is disabled until the name, namespace and targetRef are set.
After that, if there's an error when you click create, it will show an alert box below the form (see screenshot for example)

The changes also include the removal of the custom save function for DNSPolicy via the yaml editor as it isn't required (it was originally added as I thought it was needed to convert matchLabels from an object to an array, and back again at the right time for yaml saving, but that's only the case for the form view). Making this change allows the yaml editor to fall back to the built in save function for that component, which handles creation and showing validation errors.

image

TODO:

- [x] client side validation of resource namespace being non-empty, as an empty namespace causes the k8sCreate fn to POST to a slightly different url (without the namespace in it) which results in a 404, rather than a 422 with a validation error response
- [x] client side validation of resource name, as it's a standard field to have on all resources, and can easily be reused across all create views (input text type)
- [x] see if it's possible to highlight validation errors inline on specific fields in a generic way based on the 422 validation error response (or at the very least, have it for name & namespace)

### DIFF
--- a/src/components/KuadrantDNSPolicyCreatePage.tsx
+++ b/src/components/KuadrantDNSPolicyCreatePage.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import Helmet from 'react-helmet';
-import yaml from 'js-yaml';
 import {
   PageSection,
   Title,
@@ -14,8 +13,10 @@ import {
   Button,
   ExpandableSection,
   ButtonVariant,
-  Modal,
   ActionGroup,
+  AlertVariant,
+  Alert,
+  AlertGroup,
 } from '@patternfly/react-core';
 import { useTranslation } from 'react-i18next';
 import './kuadrant.css';
@@ -27,9 +28,8 @@ import HealthCheckField from './dnspolicy/HealthCheckField';
 import getModelFromResource from '../utils/getModelFromResource';
 import { Gateway } from './gateway/types';
 import GatewaySelect from './gateway/GatewaySelect';
-import { ModalHeader, ModalBody, ModalFooter } from '@patternfly/react-core/next';
 import NamespaceSelect from './namespace/NamespaceSelect';
-import { removeUndefinedFields, convertMatchLabelsArrayToObject, convertMatchLabelsObjectToArray } from '../utils/modelUtils';
+import { removeUndefinedFields, convertMatchLabelsArrayToObject } from '../utils/modelUtils';
 
 const KuadrantDNSPolicyCreatePage: React.FC = () => {
   const { t } = useTranslation('plugin__console-plugin-template');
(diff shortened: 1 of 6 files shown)


jasonmadigan:
👀

src/components/KuadrantDNSPolicyCreatePage.tsx
```diff
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import Helmet from 'react-helmet';
-import yaml from 'js-yaml';
```
jasonmadigan, line 3: Can probably drop from package.json now

src/components/KuadrantDNSPolicyCreatePage.tsx
```diff
@@ -122,39 +126,10 @@ const KuadrantDNSPolicyCreatePage: React.FC = () => {
...
-            })),
-          },
-        });
-        setHealthCheck(parsedYaml.spec.healthCheck || healthCheck);
-      }
-    } catch (error) {
-      console.error('Error parsing YAML:', error);
-    }
-    handleSubmit();
-  };
```
jasonmadigan, line 155: 🫡

src/components/KuadrantRateLimitPolicyCreatePage.tsx
```diff
@@ -316,13 +312,22 @@ const KuadrantRateLimitPolicyCreatePage: React.FC = () => {
...
-            <Alert
-              variant="info"
-              isInline
-              title="To set defaults, overrides, and more complex limits, use the YAML view."
-            />
+            <AlertGroup className="kuadrant-alert-group">
+              <Alert
+                variant="info"
+                isInline
+                title="To set defaults, overrides, and more complex limits, use the YAML view."
```
jasonmadigan, line 319: realise this is from an earlier PR, but an i18n string here would be good

jasonmadigan (commented):
Couple of small bits

src/components/KuadrantDNSPolicyCreatePage.tsx
david-martin: yes, will remove as it's not used elsewhere.
It might get re-introduced with #59

PR #73, opened 2024-09-19
Move Limit selector out to separate component

david-martin:
Small refactor to move the limit selector pieces fully out to a separate component.
Also uses the [label pattern](https://www.patternfly.org/components/label) from patternfly now.

## Before

*no limits*

image

*limits added*

image

## After

*no limits*

image

*limits added* 

image


david-martin:
Moving this back to draft to use the [labels pattern](https://www.patternfly.org/components/label/react-demos) for limits.

PR #98, opened 2024-10-10
Gh 88

R-Lawton:
Update for DNSPOLICY
Whats changed:
* Dnspolicy should be able to update via form view and yaml view
* Form view should pass values to yaml view and vice versa now not for update and create
* The new API changes for DNS policy are reflected
* Validation if required values are not filled in the create button wont show is not in the genric create/update
* Namespace selector in form is gone 
* Authpolicy now updates too
Screenshot 2024-10-14 at 11 33 31


src/components/DropdownWithKebab.tsx
```diff
@@ -38,11 +38,18 @@ const DropdownWithKebab: React.FC<DropdownWithKebabProps> = ({ obj }) => {
...
 
+  let policyType = obj.kind.toLowerCase();
 
   const onEditClick = () => {
-    history.push({
-      pathname: `/k8s/ns/${obj.metadata.namespace}/tlspolicy/name/${obj.metadata.name}/edit`,
-    })
+    if (obj.kind === 'AuthPolicy') {
+      history.push({
+        pathname: `/k8s/ns/${obj.metadata.namespace}/${obj.apiVersion.replace("/", "~")}~${obj.kind}/${obj.metadata.name}/yaml`,
```
jasonmadigan, line 46: 👍

src/components/PolicyTopologyPage.tsx
```diff
@@ -73,6 +72,9 @@ const fetchConfig = async () => {
   }
 };
 
+export const kindToAbbr = (kind: string) => {
```
jasonmadigan, line 75: import { kindToAbbr } from '../utils/modelUtils';

jasonmadigan (commented):
Few small things

jasonmadigan:
PS: apols for i18n conflicts

src/components/PolicyTopologyPage.tsx
R-Lawton: i removed the function from that file since it was now the only thing in it with the deletion of the DNS stuff but i can move it back

jasonmadigan: nah, let's leave it here for now. if we have a need for re-use again, can make another util.

Issue #90, opened 2024-10-07
Add Update to Auth policy


R-Lawton:
Following TLS add update via yaml to DNS Policy

R-Lawton:
https://github.com/Kuadrant/kuadrant-console-plugin/pull/98

PR #100, opened 2024-10-11
Wire up kebab action for edit correctly

jasonmadigan:
Resolves: https://github.com/Kuadrant/kuadrant-console-plugin/issues/90

### DIFF
--- a/src/components/DropdownWithKebab.tsx
+++ b/src/components/DropdownWithKebab.tsx
@@ -38,11 +38,10 @@ const DropdownWithKebab: React.FC<DropdownWithKebabProps> = ({ obj }) => {
     }
   };
 
-
   const onEditClick = () => {
     history.push({
-      pathname: `/k8s/ns/${obj.metadata.namespace}/tlspolicy/name/${obj.metadata.name}/edit`,
-    })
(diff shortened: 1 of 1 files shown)


src/components/DropdownWithKebab.tsx
```diff
@@ -38,11 +38,10 @@ const DropdownWithKebab: React.FC<DropdownWithKebabProps> = ({ obj }) => {
     }
   };
 
-
   const onEditClick = () => {
     history.push({
-      pathname: `/k8s/ns/${obj.metadata.namespace}/tlspolicy/name/${obj.metadata.name}/edit`,
-    })
+      pathname: `/k8s/ns/${obj.metadata.namespace}/${obj.apiVersion.replace("/", "~")}~${obj.kind}/${obj.metadata.name}/yaml`,
```
R-Lawton, line 43: this breaks the edit for the rest if the policies. I have a change currently for DNS i you want you can leave it with me and ill have it that it works for auth to

R-Lawton: /yaml is the default yaml view for all openshift resources so our view would override them which is grand for auth but no bueno for the others since they're forms

jasonmadigan: /sold

Issue #91, opened 2024-10-07
Console plugin small bugs


R-Lawton:
Small things that need to be fixed in the console plugin:

- [x] In the policies & overview view the kebab gets hidden by the page nav at the bottom ![Image] https://github.com/Kuadrant/kuadrant-console-plugin/pull/99(https://github.com/user-attachments/assets/2a5ac698-e2af-4167-8f82-b6d4347e94f5)
- [x] Change cancel to redirect back to the tab they were on in policies view not to the all policies tab #101
- [x] i18n be broken #101 
- [x] If a policy has a flip-flopping status it keeps resetting the policies view and nothing can be clicked (can't reproduce?)
- [ ] #124
- [x] #125
- [x] #127
- [x] #126
- [x] #128
- [x] #129

jasonmadigan:
Fixed three other bugs:

https://github.com/Kuadrant/kuadrant-console-plugin/pull/94

PR #101, opened 2024-10-14
small bugs

jasonmadigan:
- **Change cancel on resource create pages to navigate back, rather than go back to the all policies page.**
- **fix i18n & rename i18n source**

Resolves some issues from #91


jasonmadigan:
Related: https://github.com/Kuadrant/kuadrant-operator/pull/940/files

R-Lawton (approved):

PR #138, opened 2024-10-24
removing info box

R-Lawton:
Removing the info box as i don't think it's needed. When thinking of what to say felt like I was stating the obvious, no other resource in similar layouts has them either.


jasonmadigan:
Was thinking similar

/lgtm

PR #144, opened 2024-11-11
v1 version bumps for RLP/ALP

jasonmadigan:
Version bumps for `v1` of AP + RLP

Screenshot 2024-11-11 at 12 40 10


jasonmadigan:
cc @R-Lawton

console-extensions.json
```diff
@@ -61,7 +61,7 @@
     "properties": {
       "model": {
         "group": "kuadrant.io",
-        "version": "v1beta3",
+        "version": "v1",
```
R-Lawton, line 64: we have v1 made already you just need to remove these rather then change them

PR #165, opened 2024-11-15
replacements update

jasonmadigan:
Some additional replacements


david-martin:
/lgtm
We can revert if the replacements happen dynamically in future.

Issue #191, opened 2025-02-11
Topology: checkbox filter for show/hide resources


jasonmadigan:
**Is your feature request related to a problem? Please describe.**
Right now, the console plugin's topology view filters an arbitrary list of "noisy" resources from the topology view. It can be useful to be able to see some or all of these when debugging certain issues.

**Describe the solution you'd like**
Add a new filter checkbox area for being able to filter resource kinds to show. Define a default set (based on the existing "show) list, and then add a new [checkbox list](https://www.patternfly.org/components/forms/checkbox/#controlled) to filter what resources are shown in the graph.

## Todo

- [ ] Work out a default list (likely based on what we expect to show right now)
- [ ] Add a new filtering area to the topology UI, with a dropdown checkbox for filtering what kinds of resources to show the graph

jasonmadigan:
Based on a thought jogged from https://github.com/Kuadrant/kuadrant-operator/pull/1164

PR #195, opened 2025-02-12
Topology: add a resource filter

jasonmadigan:
Resolves https://github.com/Kuadrant/kuadrant-console-plugin/issues/191

- Added a new Resource filter with an OCP-style `Toolbar`, `ToolbarFilter` and `SelectList` for filtering specific kinds from the topology view
- Updates to the resource drop down or the picker are reflected on the graph immediately, and filters remain in place if dynamic updates for resources take place (e.g. filters remain in effect even if topology changes while on the view)
- Introduced a new `showByDefault` list of resources for minimal noise, rather than an excludes list (which was starting to become annoying to maintain. This way, new policy-machinery nodes will not automatically appear, but can be opted into by a viewer until we decide they're good to show always.

test


jasonmadigan:
I will create a follow on issue to resolve a bug stemming from this, go to resource sometimes doesn't work for some of the newer resource types that nave no mappings.  We'll need to get a dynamic resource lookup action (had something parked in a previous PR) for the `goToResource` helper.

src/components/PolicyTopologyPage.tsx
```diff
@@ -452,10 +476,53 @@ const PolicyTopologyPage: React.FC = () => {
           const { nodes, edges } = parseDotToModel(dotString);
           setParseError(null);
 
+          // Separate group nodes from normal nodes
+          const groupNodes = nodes.filter((n) => n.type === 'group');
+          const normalNodes = nodes.filter((n) => n.type !== 'group');
+
+          // Dynamically generate the list of resource types
+          const uniqueTypes = Array.from(
```
david-martin, line 484: @jasonmadigan confirming, nothing special to do here when new resources are added to the topology for them to show up in the list?
I'm thinking about ServiceMonitor & PodMonitors being added in https://github.com/Kuadrant/kuadrant-operator/pull/1164

src/components/PolicyTopologyPage.tsx
```diff
@@ -77,6 +87,24 @@ export const kindToAbbr = (kind: string) => {
   return (kind.replace(/[^A-Z]/g, '') || kind.toUpperCase()).slice(0, 4);
 };
 
+// List of resource types to show by default in the filter toolbar.
+// Only these kinds will be shown in the initial render if they exist in the parsed DOTfile
```
david-martin, line 91: This list looks reasonable, and less maintenance overhead than the inverse 
👍

src/components/PolicyTopologyPage.tsx
jasonmadigan: exactly yes. they will show up fine now, but you will need to opt-in to show them by default (via `showByDefault). Shout if there's any obs resources you feel we should show by default

david-martin:
Verified locally.
Small suggestion, for the 'Clear all filters' option.
The behaviour is to reset the filters to the default.
Would something like 'Reset filters' make more sense?

jasonmadigan:
Good point - will change, makes more sense.

jasonmadigan:
hmm, seems to be a built-in string in the `Toolbar` - not sure that can be changed from reading the docs

jasonmadigan:
@david-martin fixed

david-martin (approved):
/lgtm

PR #198, opened 2025-02-13
Fixed horizontal navigation

R-Lawton:
Fixed the horizontal navigation automatically rendering the all policies tab when a … namespace is chosen rather then staying on the previous specified navigation tab

https://github.com/user-attachments/assets/0b42ea07-c742-4d62-be38-3c7ef04a03f9

### DIFF
--- a/src/components/KuadrantPoliciesPage.tsx
+++ b/src/components/KuadrantPoliciesPage.tsx
@@ -157,6 +157,7 @@ const KuadrantPoliciesPage: React.FC = () => {
   const { ns } = useParams<{ ns: string }>();
   const [activeNamespace, setActiveNamespace] = useActiveNamespace();
   const [activePerspective] = useActivePerspective();
+  const history = useHistory();
 
   React.useEffect(() => {
     if (ns && ns !== activeNamespace) {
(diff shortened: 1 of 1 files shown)


src/components/KuadrantPoliciesPage.tsx
```diff
@@ -212,6 +213,19 @@ const KuadrantPoliciesPage: React.FC = () => {
     <PoliciesListPage resource={resources[2]} activeNamespace={activeNamespace} />
   );
 
+  const handleNamespaceChange = (activeNamespace: string) => {
+    let currentTab = '';
+    const activeTab = location.pathname.split('/').pop();
```
david-martin, line 218: I hit an issue on the 'All Policies' tab after you select an individual namespace more than once.
The url has a duplicate '/policies' in the path
e.g. `http://localhost:9000/kuadrant/ns/kuadrant-system/policies/policies`

It works ok on other tabs (I believe because of the activeTab hanging off the `/policies` like `/policies/dns`,
but not on the 'All Policies' tab because the code pops off the end of the path to determine the activeTab

src/components/KuadrantPoliciesPage.tsx
```diff
@@ -212,6 +213,23 @@ const KuadrantPoliciesPage: React.FC = () => {
...
+    }
+
+    if (activeNamespace !== '#ALL_NS#') {
+      basePath = `/kuadrant/ns/${activeNamespace}/policies`;
+    } else {
+      basePath = `/kuadrant/all-namespaces/policies`;
+    }
+    const fullPath = `${basePath}/${activeTab}`;
+
+    history.replace(fullPath);
```
david-martin, line 230: I see what you're doing here, which makes sense.
Could it be simplified (more like it was before) by setting the variable in the if/else with the tab appended?
```suggestion
    if (activeNamespace !== '#ALL_NS#') {
      basePath = `/kuadrant/ns/${activeNamespace}/policies/${activeTab}`;
    } else {
      basePath = `/kuadrant/all-namespaces/policies/${activeTab}`;
    }

    history.replace(basePath);
```

`basePath` may not make sense as the variable name though.

R-Lawton: i added the more convoluted but to stop a unnecessary / appearing at the end of the policies it has not affect on the functionality so i can remove it to make the code more readble

david-martin (approved):
/lgtm

Issue #196, opened 2025-02-12
Bug: Go to Resource in topology does not work if GVK mapping doesn't exist


jasonmadigan:
In a topology view, the `Go to resource` action doesn't work with the GVK is unknown to the plugin. Resurrect old work to "lookup" GVKs with only a K, and then navigate to these resources.

For example, with https://github.com/Kuadrant/kuadrant-console-plugin/pull/195, navigate to a `Certificate`. You'll see a failed log error message re: looking up a GVK for a certificate.

jasonmadigan:
Resolved by https://github.com/Kuadrant/kuadrant-console-plugin/pull/199

PR #199, opened 2025-02-13
196 gvk discovery

jasonmadigan:
Re: https://github.com/Kuadrant/kuadrant-console-plugin/issues/196

- Dynamically loads details about cluster resources via API discovery, so we can map Kinds to latest Groups and Versions
- Adds a new helper to warm a cache with this info on `<Topology/>` mount
- Some resources on cluster commonly shown by topology are ambiguous (e.g. ``dnsrecords.kuadrant.io` vs `dnsrecords.ingress.operator.openshift.io`)

Later on, this may morph into something more reusable as a general React hook for parts of the plugin

To verify:

- Click "Go to Resource" on any item in topology.

### DIFF
--- a/src/components/PolicyTopologyPage.tsx
+++ b/src/components/PolicyTopologyPage.tsx
@@ -50,7 +50,17 @@ import { CubesIcon, CloudUploadAltIcon, TopologyIcon, RouteIcon } from '@pattern
 
 import * as dot from 'graphlib-dot';
 import './kuadrant.css';
-import resourceGVKMapping from '../utils/latest';
+
+interface GVK {
+  group: string;
+  version: string;
+  kind: string;
+}
+let dynamicResourceGVKMapping: Record<string, GVK> = {};
+const resourceHints: Record<string, string> = {
+  Gateway: 'gateway.networking.k8s.io',
+  DNSRecord: 'kuadrant.io',
+};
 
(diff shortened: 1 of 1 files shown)


R-Lawton:
/lgtm
Got it working locally looks great nice job

src/components/PolicyTopologyPage.tsx
```diff
@@ -670,4 +687,97 @@ const PolicyTopologyPage: React.FC = () => {
...
+
+  // Helper to provide resource hinting where there could be resource ambiguity (E.g. `dnsrecords.kuadrant.io` vs `dnsrecords.ingress.operator.openshift.io`)
+  const updateMapping = (kind: string, group: string, version: string) => {
+    if (resourceHints[kind]) {
+      // Always override if the new group is the hinted group.
+      if (group === resourceHints[kind]) {
+        mapping[kind] = { group, version, kind };
+      }
+    } else if (!mapping[kind]) {
+      mapping[kind] = { group, version, kind };
```
david-martin, line 722: late to the PR, but question about this line to confirm my understanding.
This will use the first found match of the kind, unless it's one of the hinted kinds?

jasonmadigan: np: yes

PR #214, opened 2025-03-03
Add new panels to overview for gateway traffic

### LINKED ISSUE #177: Add a sub set of metrics visualisations for the platform engineer to the console plugin (Administrator perspective)
Use [QueryBrowser](https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/docs/api.md#querybrowser) to show a sub set of metrics visualisations for the platform engineer in the openshift console plugin (Administrator perspective)

The subset of visualizations will be based on the existing platform engineer grafana dashboard. There are limitations with console plugin dashboards, so we’ll need to experiment what’s possible and what gives the most value.

WHAT: Take some of the most impactful panels from the dashboard and put them on the overview page for now.

david-martin:
Closes #177

1 minor change and 1 large change.
The minor change is replacing the heading 'APIs / HTTPRoutes' with 'HTTPRoutes'

UPDATE: 13/07/25
The large change replaces the existing Gateways table with an extended columns version of it.
The additional columns show rolled up request metrics for the previous 24 hours.
If there are errors (400s, 500s), they are aggregated into a 4xx or 5xx label.
This label can be clicked for a popover that gives a breakdown by individual error code (401, 409 etc..)

See https://github.com/Kuadrant/kuadrant-console-plugin/pull/214#issuecomment-2715004542 for latest screenshots


david-martin:
Still WIP, now based off new mockups.
Lot of little bits to do still, mostly tidy up.

I think the main gateways table would be replaced by this new one that includes traffic distribution.
More focus is given to it as well as it's full width and higher up, rather than half width after policies before (getter closer to the mockups)

image

R-Lawton:
Looks great, nice job. I brought it up locally and was having a poke around as if i was a user. Two points i thought off.

1. The overview of the gateways feels like its missing something, I'm thinking something like linking off to the unhealthy gateways or more info about them. Something like, have a tooltip window like you have with the error code that lists and links off to the unhealthy gateways. 

2. if there's no current errors remove the error columns. Only show them when there is errors. The error codes are blank when there's no errors and the rate is 0.0% remove them if its not a issue right now or add something for the blank error codes like N/A etc

WDYT?

R-Lawton:
Doesnt have to be done for this pr but also thought that we should add filtering to the table to be able to filter on status of the gateways/policies or httproyutes. I can create a ticket for it

src/components/KuadrantOverviewPage.tsx
```diff
@@ -285,6 +601,107 @@ const KuadrantOverviewPage: React.FC = () => {
...
+                </CardTitle>
+              </Card>
+            </FlexItem>
+          </Flex>
+
+          <Flex className="pf-u-mt-xl">
+            <FlexItem flex={{ default: 'flex_1' }}>
+              <Card>
+                <CardTitle>
+                  <Title headingLevel="h2">{t('Gateways - Traffic Distribution')}</Title>
```
jasonmadigan, line 684: I wonder if we're missing a trick by not merging the `Gateways` & `Gateways - Traffic Distribution` cards together into just `Gateways`.

Crudely, something like (with the create gateway button inline with the resource list):
Screenshot 2025-03-21 at 12 02 02

src/components/KuadrantOverviewPage.tsx
```diff
@@ -141,6 +185,54 @@ const KuadrantOverviewPage: React.FC = () => {
...
+      title: t('Successful Requests'),
+      id: 'successfulRequests',
+      sort: 'successfulRequests',
+      transforms: [sortable],
+    },
+    {
+      title: t('Error Rate'),
+      id: 'errorRate',
+      sort: 'errorRate',
+      transforms: [sortable],
```
jasonmadigan, line 221: think Rachel mentioned similar, but a default for these if null would be cool (table looks a bit weird if no errors)

src/components/KuadrantOverviewPage.tsx
david-martin: I had a chat with Joey about this.
The first bar is meant as a 'Quick Insight' for the PE to gain a comprehensive understanding of gateways.
There are some other potential things that may land in there, like alerts.
Therefore, it might not make sense, or be scalable to have the bar combined with the gateway list.

david-martin:
> * The overview of the gateways feels like its missing something, I'm thinking something like linking off to the unhealthy gateways or more info about them. Something like, have a tooltip window like you have with the error code that lists and links off to the unhealthy gateways.

@R-Lawton that sounds reasonable.
I've passed on the comments to UX to see if there's a way to improve this.

david-martin:
> 2\. if there's no current errors remove the error columns. Only show them when there is errors. The error codes are blank when there's no errors and the rate is 0.0% remove them if its not a issue right now or add something for the blank error codes like N/A etc

I have some new mockups for how we can handle 0 errors scenario.
something like this
image

I'll need to rework a few things on the error popover to align with latest mockups, and I can include this change too.

david-martin:
Few changes:

Green Label and 'None' text for no errors
image

Rounded corners on popover.
image

If there's only 1 error code in a grouping, show that error code in the label button instead of 4xx/5xx.
image

jasonmadigan (approved):

jasonmadigan:
LGTM, needs a rebase but otherwise good to merge

david-martin:
@jasonmadigan rebased and worked in the rbac changes

PR #218, opened 2025-03-14
Gh 206

R-Lawton:
### Verification

Create an HTPasswd User
On an OCP cluster (say Hypershift), create an HTPassword User + Password. For example, call them rbactest1. This user will have no permissions or roles associated.

Login to the Console
Generate a token for oc with this user as rbactest1. Restart the local console, and verify that you can see the Access Denied warning above (along with lots of errors for any other areas of the console plugin - which we haven't yet RBACified).

Create different roles for each of the rbac scnerios below

RBAC covered:
* User doesnt have list rbac to any policy you should see:
Screenshot 2025-03-18 at 11 14 04

* User doesnt have list rbac to gateways you should see:
Screenshot 2025-03-18 at 11 14 09

* User doesnt have list rbac to httproutes you should see:
Screenshot 2025-03-18 at 11 14 13

* User only has permissions to see authpoliceis or rlp etc the other policies will be excluded from the policies view.

* User doesn't have create perms
Screenshot 2025-03-18 at 11 14 53
Screenshot 2025-03-18 at 11 14 40
Screenshot 2025-03-18 at 11 14 33

* User doesn't have edit perms 
Screenshot 2025-03-18 at 11 17 36

* user doesn't have delete perms:
Screenshot 2025-03-18 at 11 17 14

### DIFF
--- a/locales/en/plugin__kuadrant-console-plugin.json
+++ b/locales/en/plugin__kuadrant-console-plugin.json
@@ -122,7 +122,7 @@
   "There are no": "There are no",
   "There is a conflict and the resource is not accepted.": "There is a conflict and the resource is not accepted.",
   "There is a conflict on the resource.": "There is a conflict on the resource.",
-  "This view visualizes the relationships and interactions between different resources within your cluster related to Kuadrant, allowing you to explore connections between Gateways, HTTPRoutes and Kuadrant Policies.": "This view visualizes the relationships and interactions between different resources within your cluster related to Connectivity Link, allowing you to explore connections between Gateways, HTTPRoutes and Connectivity Link Policies.",
+  "This view visualizes the relationships and interactions between different resources within your cluster related to Kuadrant, allowing you to explore connections between Gateways, HTTPRoutes and Kuadrant Policies.": "This view visualizes the relationships and interactions between different resources within your cluster related to Kuadrant, allowing you to explore connections between Gateways, HTTPRoutes and Kuadrant Policies.",
   "TLS": "TLS",
   "TLSPolicy": "TLSPolicy",
   "to display - please create some.": "to display - please create some.",
--- a/locales/en/plugin__kuadrant-console-plugin.json
+++ b/locales/en/plugin__kuadrant-console-plugin.json
@@ -122,7 +122,7 @@
   "There are no": "There are no",
   "There is a conflict and the resource is not accepted.": "There is a conflict and the resource is not accepted.",
   "There is a conflict on the resource.": "There is a conflict on the resource.",
-  "This view visualizes the relationships and interactions between different resources within your cluster related to Kuadrant, allowing you to explore connections between Gateways, HTTPRoutes and Kuadrant Policies.": "This view visualizes the relationships and interactions between different resources within your cluster related to Kuadrant, allowing you to explore connections between Gateways, HTTPRoutes and Kuadrant Policies.",
+  "This view visualizes the relationships and interactions between different resources within your cluster related to Kuadrant, allowing you to explore connections between Gateways, HTTPRoutes and Kuadrant Policies.": "This view visualizes the relationships and interactions between different resources within your cluster related to Connectivity Link, allowing you to explore connections between Gateways, HTTPRoutes and Connectivity Link Policies.",
   "TLS": "TLS",
   "TLSPolicy": "TLSPolicy",
   "to display - please create some.": "to display - please create some.",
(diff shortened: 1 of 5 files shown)


jasonmadigan:
I think the root of the issue here is that the added `RBACPermissions` helper is using `checkAccess`. It does work, but as it's not a react hook, it can't really hook into React component lifecycle to avoid UI churn while loading/rendering. I think we'd be better leaning into `useAccessReview`, which is a hook.

I think a general guide to which to use would be:

use `useAccessReview` if you're in a React component
use `checkAccess` if you're outside of a React component (e.g. in some other function handler)

I wonder about the batching of checks - it may hint that we should have more discrete components, with their own useAccessReview hooks each?

david-martin:
> I think the root of the issue here is that the added `RBACPermissions` helper is using `checkAccess`

@jasonmadigan @R-Lawton I had a look through the code to understand how the perms are retrieved and map reduced.
I think the problem might be that there's truthy checks like `resourceRBAC['Gateway']?.create ?` and `resourceRBAC[obj.kind]['edit'] == true` that result in the permission error being shown if they are falsey (including not being set).
When the page is loading, those aren't set as the call to retrieve the rbac information hasn't happened yet.

The use of useEffect here in the RBACPermissions function https://github.com/Kuadrant/kuadrant-console-plugin/pull/218/files#diff-0624f29f591e504df41120fc7f8a3c54aaa37a840387ce426b9b822e8531eeadR15-R34 should be enough to ensure it fits into the react lifecycle.

Maybe a 3rd state is needed for 'loading', that precedes the other 2 states of 'permissions error' and 'loaded ok'?

jasonmadigan:
> > I think the root of the issue here is that the added `RBACPermissions` helper is using `checkAccess`
> 
> @jasonmadigan @R-Lawton I had a look through the code to understand how the perms are retrieved and map reduced. I think the problem might be that there's truthy checks like `resourceRBAC['Gateway']?.create ?` and `resourceRBAC[obj.kind]['edit'] == true` that result in the permission error being shown if they are falsey (including not being set). When the page is loading, those aren't set as the call to retrieve the rbac information hasn't happened yet.

makes sense, I think we're missing a loading state, similar to some of our other views (topology has one - https://github.com/Kuadrant/kuadrant-console-plugin/blob/main/src/components/PolicyTopologyPage.tsx#L604)
> 
> The use of useEffect here in the RBACPermissions function https://github.com/Kuadrant/kuadrant-console-plugin/pull/218/files#diff-0624f29f591e504df41120fc7f8a3c54aaa37a840387ce426b9b822e8531eeadR15-R34 should be enough to ensure it fits into the react lifecycle.

yeah I somehow missed that, that'd make this a custom hook in all but name
> 
> Maybe a 3rd state is needed for 'loading', that precedes the other 2 states of 'permissions error' and 'loaded ok'?

yep, a third state would make sense. 


still think there's value in creating more discreet components for some of the sub-views we have here (ones that spring to mind: PoliciesCard, HTTPRoutesCard, GatewaysCard, GettingStartedCard. Pushing RBAC deeper into the components feels like something we should aim for (and nudge the rbac checks closer to the components themselves rather than up here). although I'm not sure we want to do that here, or later.

jasonmadigan:
A suggestion to fix this, we should also return loading state from the hook:


```typescript
function useRBACPermissions(resources: Resource[]) {
  const [RBAC, setRBAC] = useState<Record<string, boolean> | null>(null);
  const [loading, setLoading] = useState(true);

  useEffect(() => {
      const results = await Promise.all(
        resources.flatMap(({ group, kind }) =>
          verbs.map(async (verb) => {
            const result = await checkAccess({
              group,
              resource: kind,
              verb,
            });
            return { key: `${kind}-${verb}`, isAllowed: result.status?.allowed };
          })
        )
      );
      const RBACMap = results.reduce((acc, { key, isAllowed }) => {
        acc[key] = isAllowed;
        return acc;
      setRBAC(RBACMap);
      setLoading(false);
    };

    checkRBAC();
  }, [resources]);

  return { RBAC, loading };
}
```


And then do stuff like:

```typescript
const { userRBAC, loadingUserRBAC } = useRBACPermissions(resources);
```

For the loading state, show a spinner or message:

```typescript
if (loading) {
  return <Spinner />;
}\

// or

if (loading) {
  return <div>{t('Loading ...')}</div>;
}
```


As a smaller nit, I think we should rename `RBACPermissions` to `useRBACPermissions` (seems to be a hook name convention).

jasonmadigan (approved):
LGTM

david-martin (approved):

Issue #217, opened 2025-03-10
Fix broken link in kuadrant overview


laurafitzgerald:
**Describe the bug**
Broken Docs Link in the kuadrant section of OpenShift

**To Reproduce**
1. Go to kuadrant overview section on openshift 
2. See getting Started section top left
3. click link `Configuring and deploying Gateway policies with Kuadrant
4. See 404

**Expected behavior**
Not a 404

PR #226, opened 2025-03-26
Fix broken link (Issue #217)

orlafoley:


jasonmadigan:
that guide (https://docs.kuadrant.io/latest/kuadrant-operator/doc/user-guides/full-walkthrough/secure-protect-connect-openshift/ ) is going to 404 as soon as we shift to v1.1, so I'm not sure it's a good replacement?

jasonmadigan (approved):
lgtm

downstream.js
```diff
@@ -19,7 +19,7 @@ const replacements = {
     type: 'simple',
     mappings: {
       // Order matters: specific URLs first to prevent partial matches
-      'https://docs.kuadrant.io/latest/kuadrant-operator/doc/user-guides/secure-protect-connect-single-multi-cluster/':
+      'https://docs.kuadrant.io/latest/kuadrant-operator/doc/user-guides/full-walkthrough/secure-protect-connect-openshift/':
```
david-martin, line 22: I think you'll also have to update the link in [links.ts](https://github.com/Kuadrant/kuadrant-console-plugin/blob/main/src/constants/links.ts#L22)

jasonmadigan: good spot, forgot about that one

jasonmadigan:
looks good, merging

Issue #227, opened 2025-03-26
Small fixes to Gateway traffic error codes popover


david-martin:
* [x] Change text to 'request' (singular) if there's only 1 request
* [x] Limit the max height to show 3 different error codes (if there are that many), and add a scroll bar if there are more than 3
* [x] Change the title of the popover from 'Error codes distribution' to just 'Error codes'
* [x] Sort the error requests in descending order if there's more than 1 (largest first)

![Image](https://github.com/user-attachments/assets/b2e35ca4-8789-4ea2-9e92-72436491b901)

PR #229, opened 2025-03-27
Small fixes to Gateway traffic error codes popover #227

orlafoley:
4 commits addressing the bullet points listed in issue #227

### DIFF
--- a/src/components/KuadrantOverviewPage.tsx
+++ b/src/components/KuadrantOverviewPage.tsx
@@ -384,7 +384,8 @@ const KuadrantOverviewPage: React.FC = () => {
   ): Record<string, Distribution> => {
     const key = `${obj.metadata.namespace}/${obj.metadata.name}-istio`;
     const codes = totalRequestsByGateway[key]?.codes ?? {};
-    const filteredCodes = Object.entries(codes).filter(([code]) => code.startsWith(prefix));
+    const filteredCodes = Object.entries(codes).filter(([code]) => code.startsWith(prefix))
+    .sort(([, count1], [, count2]) => count2 - count1);
 
     const total = filteredCodes.reduce((sum, [, count]) => sum + count, 0);
 
--- a/src/components/KuadrantOverviewPage.tsx
+++ b/src/components/KuadrantOverviewPage.tsx
@@ -430,7 +430,7 @@ const KuadrantOverviewPage: React.FC = () => {
                           <strong>Code: {code}</strong>
                         </FlexItem>
                         <FlexItem align={{ default: 'alignRight' }}>
-                          {dist.total.toFixed(0)} requests
+                          {dist.total.toFixed(0) === '1' ? '1 request' : `${dist.total.toFixed(0)} requests`}
                         </FlexItem>
                       </Flex>
                     }
--- a/src/components/KuadrantOverviewPage.tsx
+++ b/src/components/KuadrantOverviewPage.tsx
@@ -388,7 +388,7 @@ const KuadrantOverviewPage: React.FC = () => {
 
     const total = filteredCodes.reduce((sum, [, count]) => sum + count, 0);
 
-    let distribution: Array<[string, Distribution]> = [];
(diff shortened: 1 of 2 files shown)


david-martin (changes requested):
Thanks for getting the PR together @orlafoley 

The failing DCO CI check is due to commits not being signed. You can find more info on that by clicking the failing check, or [here](https://github.com/Kuadrant/kuadrant-console-plugin/pull/229/checks?check_run_id=39516516949).

The Lint CI failure should be easily fixed by running `yarn lint`.

In general, the changes look good.
The max height though seems to be behaving oddly for me.

The scrollbar is showing even with just 2 error codes visible, but the bottom part of the popover pointing to the error code label is not visible.
image

Then when you scroll down, the point is visible.
image

From looking back at the mockups (internal mockups doc), I think the scrollbar would sit better on just the error codes and bars. What do you think?
image

david-martin:
It doesn't seem to be forcing the max height for me
image

To test this, I hardcoded extra some distributions in the page
```tsx
    const distribution: Record<string, Distribution> = {
      '429': { total: 60, percent: 60 },
      '404': { total: 39, percent: 39 },
      '403': { total: 39, percent: 39 },
      '402': { total: 39, percent: 39 },
      '401': { total: 1, percent: 1 },
    };
```

I tried a few variations of css but couldn't figure out why the max height wasn't being enforced, even when setting `!important` on it. This might need some trial and error to figure out.

What do think of setting the overflow to `auto` as well so the scrollbar only shows when needed?

src/components/KuadrantOverviewPage.tsx
```diff
@@ -409,37 +411,41 @@ const KuadrantOverviewPage: React.FC = () => {
...
-                        </FlexItem>
-                      </Flex>
-                    }
-                    measureLocation={ProgressMeasureLocation.outside}
-                  />
-                  <Divider style={{ margin: '12px 0' }} />
-                </div>
-              );
-            })}
+            <div className="custom-rounded-popover-codes">
```
R-Lawton, line 420:
```suggestion
          <div style={{ maxHeight: '300px', overflowY: 'auto' }}>
```
I wasn't seeing the max height kicking in either had a quick look to see if i could figure it out. Found that adding a inline style over a custom CSS class works, why? no clue haven't looked to deep into it but it works.

src/components/KuadrantOverviewPage.tsx
```diff
@@ -384,7 +384,9 @@ const KuadrantOverviewPage: React.FC = () => {
   ): Record<string, Distribution> => {
     const key = `${obj.metadata.namespace}/${obj.metadata.name}-istio`;
     const codes = totalRequestsByGateway[key]?.codes ?? {};
-    const filteredCodes = Object.entries(codes).filter(([code]) => code.startsWith(prefix));
+    const filteredCodes = Object.entries(codes)
+      .filter(([code]) => code.startsWith(prefix))
+      .sort(([, count1], [, count2]) => count2 - count1);
```
R-Lawton, line 389: the sorting isnt showing in decending order 
Screenshot 2025-04-08 at 15 19 54

R-Lawton:
Hey Orla, Noticed two things, let me know if you need any help with getting it over the line happy to help

R-Lawton (changes requested):
This should fix the issues with the indices

R-Lawton:
I cant suggest as you havent made changes to this part of the code in this pr but with the new way of doing the error codes we can modify the const distribution into the same style Array and remove the object.enteries on the sorted distribution so its the same return type without the need for the extra steps

R-Lawton:
The sorting looks good, the max height isnt working anymore for some reason i don't see why nothing much has changed. Ill take a look to see if i can figure out whats changed between my suggestion and now, if you want to do the same

R-Lawton:
Max height looks good but in the status 500 popover im seeing status 400s intermixed
Screenshot 2025-04-14 at 10 32 59

PR #238, opened 2025-05-22
bump 0.2.0-dev

eguzki:


jasonmadigan (approved):

eguzki:
Bumping to the next minor version. Hopefully this is fine. We can always cut a new patch release  (0.1.4 would be the next one) if needed).

Issue #235, opened 2025-05-20
Fix bug: Gateway status not updating dynamically without page refresh


emmaaroche:
**Describe the bug**
When a gateway transitions to an "enforced" state, the "Healthy Gateways" count on the overview page does not update automatically. The correct count is only reflected after a manual page refresh. Same with status for  "Unhealthy Gateways".

**To Reproduce**

1. Navigate to the overview page in the console plugin where gateways are listed
2. Trigger a change that causes a gateway to become enforced
3. Observe the "Healthy Gateways" or "Unhealthy Gateways" count
4. Note that it does not update until the page is refreshed manually

**Expected behavior**
The "Healthy Gateways" and "Unhealthy Gateways" count should update automatically without requiring a page refresh, reflecting the gateway status in real time.

**Screenshots**
Screenshot shows an enforced gateway while status is still 0 for "Healthy Gateways"
<img width="1401" alt="Image" src="https://github.com/user-attachments/assets/8ae9b0a2-b593-43e0-ab11-2e295df778f9" />

**Additional context**
This behavior was observed during manual UI testing on a test OpenShift cluster using RHCL v1.1.

jasonmadigan:
thanks @emmaaroche, definitely a bug.

PR #239, opened 2025-05-30
Gateway Bug #235

orlafoley:
Updated `KuadrantOverviewPage.tsx` to refresh gateway counts every 5 seconds rather than needing a manual refresh for an update

### DIFF
--- a/src/components/KuadrantOverviewPage.tsx
+++ b/src/components/KuadrantOverviewPage.tsx
@@ -46,7 +46,7 @@ import {
   usePrometheusPoll,
   PrometheusEndpoint,
   K8sResourceCommon,
-  k8sList,
+  useK8sWatchResource,
   GreenCheckCircleIcon,
   YellowExclamationTriangleIcon,
   TableData,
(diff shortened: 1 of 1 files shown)


R-Lawton:
can you give more context on this or link to the issue i dont know what the bug is. Verification steps would be useful to

R-Lawton:
The values do get synced after a few seconds but its looks like you've created a delay with the new methodology thats last for 5s the value do change after that if new gateways change health etc but we dont want that intial delay
Screenshot 2025-06-11 at 15 11 29

src/components/KuadrantOverviewPage.tsx
```diff
@@ -554,8 +562,14 @@ const KuadrantOverviewPage: React.FC = () => {
       }
     };
 
+    // Call the function once to prevent the initial 5 second delay
     fetchGateways();
-  }, []);
+
+    // Check if there's a gateway change every 5 seconds, run the fetch gateway function periodically
+    const interval = setInterval(fetchGateways, 5000);
```
jasonmadigan, line 569: I think a better fix here would be to switch to different pattern to fetch new data.

`k8sList` is a one-time fetch, we'd be better of replacing with `useK8sWatchResource`, which will maintain a WS watch for changes in Gateway resources, and update things dynamically. You'll see examples of this usage elsewhere in the policies and other components - for example in `GatewaySelect.tsx`, `ResourceList.tsx`.

instead of `useState`, switch to `useMemo` (similar to  the patterns elsewhere like `GatewaySelect.tsx`). 

`useState` is sorta like creating variable that describes stuff, but you have to take care of all the details on how to update. `useMemo` is a better fit here because the watch hook is already managing the data updates for us - we just need to transform/filter that data when it changes. with `useState`, we'd need to write a `useEffect` hook to copy watched data back and forth on updates, which is a bit tedious.

jasonmadigan (changes requested):

jasonmadigan (changes requested):
much better, thanks @orlafoley. Some small remaining comments - there's a lint error with a build too, but it's likely the commented out block.

jasonmadigan:
LGTM - thanks @orlafoley

jasonmadigan (approved):

Issue #254, opened 2025-06-20
Visual coloring glitches on Openshift 4.19


azgabur:
**Describe the bug**
While testing on ocp419 I noticed some visual bugs in _Overview_ page of the plugin.
- Coloring on texts in _Getting started resources_ section seems to not apply
- Background coloring in lists of resources seem to not match the new ocp419 background color

**To Reproduce**
1. Create any Gateway, Policy
2. See mentioned issues

**Screenshots**

![Image](https://github.com/user-attachments/assets/50c501ea-ff93-476e-8084-474feec5d661)

![Image](https://github.com/user-attachments/assets/1bee5649-cde1-4e22-ada5-d757da1471e1)

**Additional context**
Version of Kuadrant: 1.2.0

PR #256, opened 2025-06-27
DRAFT: Visual coloring glitches on Plugin Overview Page

tommyc2:
Ref: https://github.com/Kuadrant/kuadrant-console-plugin/issues/254

### What is the bug?
While testing on OCP 4.19, user reported some visual bugs in Overview page of the plugin.

- Coloring on texts in Getting started resources section seems to not apply
- Background coloring in lists of resources seem to not match the new OCP 4.19 background color

### To reproduce
1. Visit Kuadrant Overview Page using side menu
2. Note the differences using OCP 4.17/4.18 overview pages as comparison

### It was
image
image

### It is now
image
image
image

### DIFF
--- a/src/components/KuadrantOverviewPage.tsx
+++ b/src/components/KuadrantOverviewPage.tsx
@@ -655,7 +655,7 @@ const KuadrantOverviewPage: React.FC = () => {
               </Card>
             )}
 
-            <Flex className="pf-u-mt-xl">
+            <Flex className="pf-v5-u-mt-xl">
               <FlexItem flex={{ default: 'flex_1' }}>
                 <Card>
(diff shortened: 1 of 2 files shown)


jasonmadigan:
Yeah, I think smaller issues like that we can live with while we move to PF6

jasonmadigan:
Looks pretty good @tommyc2 thanks a lot. One small tweak: think we can we get our margin between the panels back, similar to before for the cards? crude screenshot to show what I mean :)

Screenshot 2025-07-07 at 12 01 05

jasonmadigan (approved):
LGTM, cheers @tommyc2

PR #258, opened 2025-07-04
Update base image for Dockerfile to UBI 9 standard

tommyc2:
Updating the base image used in the Dockerfile to the correct UBI 9 version.

`registry.access.redhat.com/ubi9/ubi:latest` is the recommended path for the standard UBI 9 image.

`registry.access.redhat.com/ubi9:latest` relies more on the registry default within the ubi9 space


R-Lawton:
👀

Issue #14, opened 2024-07-29
Flag Kuadrant CR status/conditions for dependencies in the plugin


jasonmadigan:
**What**

In console, we want to provide some assistance to users when they've installed Kuadrant. In particular, we want to flag if dependencies are missing post-install. The latest `kuadrant` CR now has status/conditions about a number of our dependencies, that can be used as an API to build this view. See: https://github.com/Kuadrant/kuadrant-operator/pull/994

This will be useful in terms of visualising the "state" of Kuadrant. We should look to surface this on the "Overview" Page.

david-martin:
> * There may be a mechanism in Operator Hub already to flag operator dependencies already

Visual warnings, calling out dependencies & linking to additional docs as a minimum.
See this example for what I mean:

<img width="853" alt="image" src="https://github.com/user-attachments/assets/592b84e4-1357-4d60-9c49-ce6b6fe450d1">

jasonmadigan:
Interdependency between this and https://github.com/Kuadrant/console-plugin/issues/20

jasonmadigan:
Shifted to 1.1

jasonmadigan:
https://github.com/Kuadrant/kuadrant-operator/pull/994

jasonmadigan:
I think it is, it needs designs, but we now have an API to actually build this against (previously, we did not). Will add a `needs design` label.

Joeyyubo:
I got a good documentation on this:https://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/surfacing-operator-dependencies/

I guess it may be a mechanism in Operator Hub already to flag operator dependencies already. I'm looking for the latest version.

david-martin:
@jasonmadigan looking at the fix in https://github.com/Kuadrant/kuadrant-console-plugin/pull/259, I think this issue has morphed a little to the point where there's 2 levels at which dependency help can be given.
The original description and comments were more about at install time of the operator if I'm not mistaken,
while the approach in the linked PR is more inline with the updated issue description above (probably on purpose). That is, after installation and creation of the Kuarant CR, are there any detected missing dependencies (like GWAPI CRDs or Isito).

In the linked doc from Joey, https://openshift.github.io/openshift-origin-design/designs/administrator/future-openshift/surfacing-operator-dependencies/, pre and post installation is called out as well, though with different suggestions fro post that the approach we're taking. I can see value in our current approach though, given we have a plugin overview page.

[...]

PR #259, opened 2025-07-10
[WIP] Add Alert on Overview page about Kuadrant CR Status

tommyc2:
WIP FYI

### DIFF
--- a/src/components/KuadrantOverviewPage.tsx
+++ b/src/components/KuadrantOverviewPage.tsx
@@ -59,6 +59,7 @@ import resourceGVKMapping from '../utils/latest';
 import { useHistory } from 'react-router-dom';
 import useAccessReviews from '../utils/resourceRBAC';
 import { getResourceNameFromKind } from '../utils/getModelFromResource';
+import { KuadrantStatusAlert } from './KuadrantStatusAlert';
 
 export type MenuToggleElement = HTMLDivElement | HTMLButtonElement;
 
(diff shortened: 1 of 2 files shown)


tommyc2:
https://github.com/Kuadrant/kuadrant-console-plugin/issues/14

tommyc2:
> This is a great start @tommyc2
> 
> Couple of thoughts suggestions:
> 
> * Can we look at a smaller area to flag status? Maybe instead of an `<Alert/>`, we could look at a smaller icon + status only thing to the top right of that card? Something some to the top-right of the card like like: `<Icon> OK`. And then maybe a PopOver with that icon for a more "in-depth" status
> * I think we're only looking at surfacing one condition - ideally we'd be able to show all of these in the popover? Maybe a line for each status condition in the popover (I'll let you look at what you think might look best here)
> * The overview card can be hidden (see the triple dots) - I'm thinking we could perhaps drop that functionality if status is going to be in that card. It would be a shame if users could dismiss that helpful info longer term.

Thanks for the suggestions Jason 🙌

I’ll look at applying those recommendations

tommyc2:
> FYI @tommyc2 @jasonmadigan
> 
> For displaying alerts, here is a PF design recommendation of "Toast alert" which keeps the design style consistent with OCP console. https://www.patternfly.org/components/alert/html-demos#toast https://www.patternfly.org/components/alert/design-guidelines#toast-alerts (PF6)
> 
> For multiple status conditions, If there are multiple alerts to stack and fill the screen, "alert group" maybe a good option. When applying "alert group" , a threshold of 3 alerts is recommended to display before showing the footer. https://www.patternfly.org/components/alert/design-guidelines/#alert-groups

Yeah but I think what Jason is saying is that it would be nice to have that information available long term (e.g. not a temporary alert)

tommyc2:
@jasonmadigan yeah i like that approach, i had fixed it by specifiying the namespace just there but then realised it can be anywhere. Ill test that now

jasonmadigan (approved):

PR #267, opened 2025-07-15
Patternfly 6 upgrade

jasonmadigan:
- Upgrades libraries and CSS from PatternFly 5 to PatternFly 6
- Resolves some bugs with state and topology filtering


R-Lawton:
Few things i noticed 
* The http route and policies card on the overview used to be side by side now there one above the other.

pf6:
Screenshot 2025-07-17 at 12 00 27

pf5:
Screenshot 2025-07-17 at 12 00 41


* The tab headings in the policies the font is quite small compared to before
pf6
Screenshot 2025-07-17 at 11 57 49


* This is more of a nit pick but the values in all the tables seem to be slightly higher in the rows and not centerd compared to pf5

pf5:
Screenshot 2025-07-17 at 12 06 00

pf6:
Screenshot 2025-07-17 at 12 05 48

jasonmadigan:
Thanks @R-Lawton 

> The http route and policies card on the overview used to be side by side now there one above the other.

Fixed

> The tab headings in the policies the font is quite small compared to before

Seems to be a general PF6 typography change, so I'll probably keep as-is for the sake of general console consistency. See `/settings/cluster`:

Screenshot 2025-07-21 at 11 13 51

> This is more of a nit pick but the values in all the tables seem to be slightly higher in the rows and not centerd compared to pf5

looks odd to me too, addressed now

R-Lawton (approved):
/lgtm verified locally

PR #268, opened 2025-07-22
Some post PF6 fixes

jasonmadigan:
- missing i18n keys, fixed wrappers on /policies for httproutes + gateways
- Preserve topology connections when filtering out intermediate nodes (fixes a bug where `listeners` being hidden removed connections between `Gateway` and `HTTPRoute` resources)
- isolate the `usePrometheusPoll()` re-renders on the overview page, to prevent frequent re-rendering elsewhere


R-Lawton:
👀

PR #276, opened 2025-09-23
fix dns/tls policy edit to use forms

Patryk-Stefanski:
Searching for TLS/DNS policies and clicking on three dots should now bring you to the custom form page when editing.

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -120,6 +120,28 @@
     }
   },
   {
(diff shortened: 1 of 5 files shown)


jasonmadigan:
Screenshot 2025-09-24 at 11 42 00

jasonmadigan:
Looks fine, padding issue in the section header was an existing issue. Can fix now or raise a bug, don't mind which.

PR #346, opened 2026-03-27
Proposal (Design doc) for how we should handle rbac in the OCP dev po…

R-Lawton:
…rtal


R-Lawton:
close in favour off https://github.com/Kuadrant/kuadrant-console-plugin/blob/main/docs/designs/2026-03-26-api-management-rbac-design.md

PR #366, opened 2026-04-14
API Product custom list

### LINKED ISSUE #316: API Products List page
#WHAT: 
Create a list view for APIProducts. The user should be able to see a list of APIProducts depending on their specific Role/RBAC.

Final designs are being worked on, but may look something like 

Image

LINK TO FINAL DESIGN: https://www.figma.com/design/nDsKwCU06OyhXH7bvOHxap/-Latest--RHCL-dev-portal-in-OCP?node-id=40-14335&m=draw

didierofrivia:
Partially closes https://github.com/Kuadrant/kuadrant-console-plugin/issues/316

This PR continues the work done in https://github.com/Kuadrant/kuadrant-console-plugin/pull/349, introducing the custom columns and filters for the API Product list.

Screenshot 2026-04-29 at 16 21 42

Screenshot 2026-04-29 at 16 23 02

TODO in a follow up PR:

1. Add Tooltips
2. Status filter should be [PF6-toolbar with filter](https://www.patternfly.org/components/toolbar#with-filters)
3. Search box should be [PF6-Attribute search](https://www.patternfly.org/patterns/filters/#attribute-search)
4. Add HTTPRoute filter

Captured in https://github.com/Kuadrant/kuadrant-console-plugin/issues/398

## Summary by CodeRabbit

* **New Features**
  * Dedicated API Products list page with searchable table, status filtering, pagination, PlanPolicy lookup and enhanced columns (Route, Tags, Version).

* **Bug Fixes / UX**
  * Improved per-resource loading/error handling and clearer empty/filtered-state messaging.

* **Documentation / Localisation**
  * Added English UI strings for API Products list states and new labels.

* **Navigation**
  * Console navigation entries updated to route via the API Product resource.

* **Tests**
  * New end-to-end fixtures, RBAC updates and E2E tests covering listing, filtering and UI states.


console-extensions.json
```diff
@@ -255,10 +255,11 @@
     }
   },
   {
-    "type": "console.navigation/resource-ns",
```
eguzki: Why moving from a dynamic, namespace-aware navigation to a static link that always opens the all-namespaces view?

didierofrivia: I failed to make it work with the custom view, when it was dynamic it was always rendering the autogenerated one :S, what would be the correct way of rendering custom here?

src/components/apiproduct/APIProductsListPage.tsx
```diff
@@ -20,18 +55,502 @@ const APIProductsListPage: React.FC = () => {
...
+    const map = new Map<string, PlanPolicy>();
+    if (planPoliciesLoaded && planPolicies) {
+      planPolicies.forEach((policy) => {
+        const targetRef = policy.spec?.targetRef;
+        if (targetRef && targetRef.kind === 'HTTPRoute' && targetRef.name) {
+          // Use the policy's namespace if targetRef.namespace is not specified
+          const targetNamespace = targetRef.namespace || policy.metadata?.namespace;
+          const key = `${targetNamespace}/${targetRef.name}`;
+          // Store first matching policy (could be extended to handle multiple)
+          if (!map.has(key)) {
```
eguzki, line 108: nit: what if there are two plan policies targeting the same httproute?

didierofrivia: Then we might want to consult to the future UXD team to review the designs, since it was meant to be just the one showing. Should I leave it as it is today or remove it?

console-extensions.json
didierofrivia: Ok, this is fixed, we need to add the definition above of the list page and its component:

```json
{
    "type": "console.page/resource/list",
    "properties": {
      "model": {
        "group": "devportal.kuadrant.io",
        "version": "v1alpha1",
        "kind": "APIProduct"
      },
      "component": { "$codeRef": "APIProductsListPage" }
    }
  },
```

src/components/apiproduct/APIProductsListPage.tsx
eguzki: I think you can leave it as it is. We can address that later if needed.

R-Lawton:
👀

R-Lawton:
Do we want to add some playwright tests for this new work?

didierofrivia:
@R-Lawton totally, will add some :+1:

src/components/apiproduct/APIProductsListPage.tsx
```diff
@@ -1,37 +1,563 @@
...
+  });
+
+  // Watch PlanPolicy resources to link them to APIProducts
+  const [planPolicies, planPoliciesLoaded, planPoliciesLoadError] = useK8sWatchResource<
+    PlanPolicy[]
+  >({
+    groupVersionKind: RESOURCES.PlanPolicy.gvk,
+    namespace: activeNamespace === '#ALL_NS#' ? undefined : activeNamespace,
+    isList: true,
+  });
```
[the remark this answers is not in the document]
didierofrivia: No

console-extensions.json
```diff
@@ -111,6 +111,17 @@
       "component": { "$codeRef": "APIProductsListPage" }
     }
   },
+  {
+    "type": "console.page/resource/list",
```
R-Lawton, line 116: i dont think we need the old route above this for apiproducts list anymore i dont think its linked to anything

didierofrivia: Do you reckon?, I thought that we still needed a route for _all-namepaces_ like we do have in the policies list here: https://github.com/Kuadrant/kuadrant-console-plugin/blob/052c80c102ccc13f1c5c156df72aab1e6024440a/console-extensions.json#L79-L91 
If that too is incorrect I could update here or create an issue

R-Lawton: Your essentially using both custom routing and also standard k8s routing. console.navigation/resource-ns generates standard k8s paths. But the components handleNamespaceChange navigate somewhere custom.  The rest of the plugin uses custom routes and custom nav. Lets leave and have a discussion later as this might change depending on the final resting place of the apiproduct list

R-Lawton: No actually the more i think of it the more i think we should pick one. The rest of the console goes with custom routing so i think we shouldnt move away from that. So that means changing the nav to use the custom routing and not having multiple routing options. Does that make sense?

didierofrivia: We could fix it later unifying the routing in a following issue. I've removed the APIProductList ones in this PR

eguzki:
in #316 section 5 is about Tooltips. Those are not implemented. Any plan for those?

eguzki:
In #316, in the toolbar, the status filter is not a [PF6-toolbar with filter](https://www.patternfly.org/components/toolbar#with-filters). Multiple status cannot be selected.

eguzki:
The "other" filter (the one which now currently supports name and namespace), does not allow filtering by route as spec in #316. The search component is not [PF6-Attribute search](https://www.patternfly.org/patterns/filters/#attribute-search) as defined in #316.

src/components/apiproduct/APIProductsListPage.tsx
```diff
@@ -37,27 +93,501 @@ const APIProductsListPage: React.FC = () => {
...
+        activeColumnIDs: Set<string>,
+      ) => {
+        return (
+          <TableData key={column.id} id={column.id} activeColumnIDs={activeColumnIDs}>
+            <Timestamp timestamp={resource.metadata?.creationTimestamp} />
+          </TableData>
+        );
+      },
+    }),
+    [planPolicyMap],
```
eguzki: Missing translation tependency in renderers Memo

e2e/tests/apiproduct-crud.spec.ts
```diff
@@ -396,7 +408,8 @@ test.describe('APIProduct CRUD Operations', () => {
     await expect(saveButton).toBeVisible();
   });
 
-  test('should delete APIProduct with confirmation', async ({ page }) => {
+  // Skipping since due to the new columns in the list and the resolution of the test screen, the kebab menu is hidden
```
eguzki, line 411: I can still see the kebab menu 🤔 

In any case, if the test is no longer valid, then 🔥 to it.

didierofrivia:
@eguzki some of the things in the issue description was supposed to be adjusted in a thread with UXD, was pending some questions I had... since that's not gonna happen, I propose to add  them as suggested in the original issue in a follow up PR. Would that be OK? So this branch doesn't grow more than it is as of today. Will add the TODOs in the description and not close the issue

eguzki (commented):
Overall looks good. Some questions already left regarding some spec in #316 not being addressed here. 

The page doesn't reset filter state when activeNamespace changes. Consider resetting filters.

e2e/tests/apiproduct-crud.spec.ts
didierofrivia: It's hidden in the test suite because of its settings, not when rendered manually in a browser

eguzki:
> @eguzki some of the things in the issue description was supposed to be adjusted in a thread with UXD, was pending some questions I had... since that's not gonna happen, I propose to add them as suggested in the original issue in a follow up PR. Would that be OK? So this branch doesn't grow more than it is as of today. Will add the TODOs in the description and not close the issue

Happy to skip those in this PR. But make sure those requests are written down in some github issue or when #316 closes, they will be gone.

eguzki (approved):
LGTM

PR #384, opened 2026-05-01
Add Target sub-tab to Policy detail views

### LINKED ISSUE #269: For Policy details views, add targetRef sub-view
For `Gateway` and `HTTPRoutes`, we have a subview added as decoration which shows attached policies for these resources.

For the policy detail views themselves however, we don't show their relationship back to their targetRef (be it an `HTTPRoute` or `Gateway`).

Let's make these similar:

- Add a new sub-view to our resource details views, showing associated targetRef(s)

shiavm006:
Mirrors the existing "Policies" tab on Gateway and HTTPRoute by adding a "Target" tab to AuthPolicy, RateLimitPolicy, DNSPolicy, and TLSPolicy detail pages. The tab reads spec.targetRef from the policy and renders the target Gateway/HTTPRoute with a ResourceLink and status, completing the bidirectional view between policies and the resources they attach to.

Closes #269

## Summary by CodeRabbit

* **New Features**
  * Added navigation tabs for Kuadrant policy types and devportal APIProduct.
  * Added a Target Reference page showing a policy’s target name, type, namespace and status, with loading, error and empty states.

* **Documentation**
  * Added user-facing strings for target reference labels, loading/errors and validation.

* **Chores**
  * Registered the Target Reference page for dynamic loading by the console.


shiavm006:
cc @eguzki  @emmaaroche  could u review it once would find it great if there is any mistake or enhancement to work on

shiavm006:
hey @eguzki  could u check out the pr review it

shiavm006:
hey @eguzki  could u just check the pr

src/components/PolicyTargetRefPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+            </Title>
+          }
+          icon={SearchIcon}
+        >
+          <EmptyStateBody>{t('This policy does not declare a spec.targetRef.')}</EmptyStateBody>
+        </EmptyState>
+      );
+    }
+    return (
+      <VirtualizedTable<TargetRef>
```
eguzki: Using VirtualizedTable for a single row is heavy-handed. Consider using PatternFly's DescriptionList or DataList for better performance and simpler.

The target is always one and, even if there are enhancement proposal for multiple target refs, it's not planned yet for kuadrant.

console-extensions.json
```diff
@@ -44,6 +44,38 @@
...
+  },
+  {
+    "type": "console.tab/horizontalNav",
+    "properties": {
+      "model": { "group": "kuadrant.io", "version": "v1", "kind": "DNSPolicy" },
+      "page": { "name": "Target", "href": "targetref" },
+      "component": { "$codeRef": "PolicyTargetRefPage" }
+    }
+  },
+  {
```
eguzki, line 71: PlanPolicies and TokenRatelimitPolicies are missing

APIPRoduct CRD also has a target reference.

src/components/PolicyTargetRefPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+        columns={columns}
+        Row={TargetRow}
+      />
+    );
+  };
+
+  return (
+    <>
+      <Helmet>
+        <title data-test="example-page-title">{t('Target Reference')}</title>
```
eguzki: `data-test="example-page-title"` something more specific?

eguzki (commented):
We need some UX design for this view, but this is a very good starting point. 

Left some comments

src/components/PolicyTargetRefPage.tsx
shiavm006: agree it doesn't make sense for a single row swapped to `DescriptionList`

console-extensions.json
shiavm006: added all three — `PlanPolicy`, `TokenRateLimitPolicy`, and `APIProduct`. also added `OIDCPolicy` for consistency since it also has a `targetRef`. we can remove if you'd rather keep the scope tighter

src/components/PolicyTargetRefPage.tsx
shiavm006: renamed to `policy-target-ref-page-title`

src/components/PolicyTargetRefPage.tsx
```diff
@@ -0,0 +1,169 @@
...
+type PolicyResource = K8sResourceCommon & {
+  spec?: {
+    targetRef?: TargetRef;
+  };
+};
+
+// extract the kind segment from a console URL like
+// /k8s/ns/<ns>/<group>~v<ver>~<Kind>/<name>/...
+// version may be v1, v1alpha1, v2beta3 etc.
+const extractKindFromURL = (pathname: string): string | null => {
```
eguzki: Isn't this method exist in nameFromPath.ts ? Is this doing something else?

src/components/PolicyTargetRefPage.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const targetWatch =
+    targetMeta && targetRef && targetNamespace
+      ? {
+          groupVersionKind: targetMeta.gvk,
+          namespace: targetNamespace,
+          name: targetRef.name,
+          isList: false,
+        }
+      : null;
+  const [target, targetLoaded, targetLoadError] =
```
eguzki, line 83: if the target has not been loaded, what is the view showing?

eguzki (changes requested):
There is something wrong with the `getStatusLabel`. The page is broken. I cleared cache and rebuilt and the issue still persists. 

Image

shiavm006:
i went through it `getStatusLabel` calls `useTranslation` inside a non hook function and i was calling it conditionally if i am not wrong this were i am violating the rules

src/components/PolicyTargetRefPage.tsx
shiavm006: that one returns the name from after the GVK, mine returns the kind from inside it. moving both to nameFromPath.ts

eguzki:
Commit messages are incomprehensible

eguzki:
Address coderabbitai reviews

shiavm006:
@eguzki addressed CodeRabbit's URL-parsing edge case also rewrote the commit message to be cleaner. let me know if anything else needs work.

eguzki:
I can still see one commit message `...`

shiavm006:
oh sorry just a sec

shiavm006:
@eguzki rebased into a single clean commit.

shiavm006:
@eguzki same e2e-rbac flake  kuadrant CR doesn't reach ready after the operator deploys.

src/components/PolicyTargetRefPage.tsx
```diff
@@ -0,0 +1,181 @@
...
+type TargetStatusProps = {
+  target: K8sResourceCommon | undefined;
+  loaded: boolean;
+  loadError: unknown;
+};
+
+// getStatusLabel calls useTranslation internally, so it must run on every
+// render to keep the hook order stable. Calling it unconditionally with a
+// placeholder when target isn't loaded yet preserves Rules of Hooks.
+const PLACEHOLDER_TARGET = {
```
eguzki: The problem is that the getStatusLabel utility function shouldn't be calling hooks internally.

Refactor getStatusLabel to accept t as a parameter instead of calling useTranslation internally. I tried it and works.

Image

> Note: leave in your PR a screenshot of the UI your building, That helps the reviewing

wdyt?

shiavm006: like i also think utility shouldn't have hooks inside it hides the dependency from callers.
i mean passing t as a param fixes it at the source rather than working around it

shiavm006:
@eguzki  did the refactor getStatusLabel(t, obj) i feels like it works correct now 
what u think 
Screenshot 2026-05-08 at 2 08 25 AM
Screenshot 2026-05-08 at 2 05 04 AM
Screenshot 2026-05-08 at 2 03 50 AM

eguzki (approved):
LGTM

Nice one

PR #392, opened 2026-05-03
feat: add form and yaml edit views for kuadrant policy resources

Aryanburnwal05:
What
Adds a custom edit experience for Kuadrant policy resources with both Form and YAML views.

Changes

1. Introduced a custom edit route for Kuadrant policy resources
2. Added a new edit page component supporting:
     - Form-based editing (default)
      - YAML editing using OpenShift's ResourceYAMLEditor
3. Implemented CRD-aware form fields (e.g. targetRef configuration)
4. Added a safety check to ensure the custom editor only applies to Kuadrant policies
5. Added an ErrorBoundary to fall back to YAML view if the form fails
6. Updated navigation to route policy edits through the custom editor

Why

The default YAML editor is not always ideal for users. This provides a more structured and user-friendly editing experience while still allowing full YAML access when needed.

Notes

1. Applies only to Kuadrant policy CRDs
2. Non-policy resources fall back to the default OpenShift editor
3. YAML view remains available as a fallback at all times
4. Form implementation covers key fields and can be extended further if needed

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -130,6 +130,14 @@
     }
   },
   {
+    "type": "console.page/route",
+    "properties": {
+      "exact": true,
+      "path": "/k8s/ns/:ns/:kind/name/:name/edit",
+      "component": { "$codeRef": "KuadrantPolicyEditPage" }
+    }
+  },
+  {
     "type": "console.resource/create",
     "properties": {
       "model": {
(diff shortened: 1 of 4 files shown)


R-Lawton:
Thank you for the interest, but this is work specifically for the LFX mentorship program, which is still in the process of picking mentees.  If you'd like to take part in the mentorship, please go through the right channels

PR #393, opened 2026-05-03
feat: add form-based create and edit views for RateLimitPolicy

Aryanburnwal05:
What
This adds form-based create and edit views for RateLimitPolicy resources, along with a YAML editor option.

Changes

1. Added new create and edit pages for RateLimitPolicy
2. Introduced a form view for common fields like targetRef and basic rate limit configuration
3. Added YAML editing using OpenShift's ResourceYAMLEditor
4. Added a toggle to switch between Form (default) and YAML views
5. Wired up a custom edit route for RateLimitPolicy resources
6. Updated kebab actions to use the new edit flow

Why
Editing RateLimitPolicy purely through YAML can be difficult for common use cases. This adds a simpler form-based experience while still allowing full YAML control when needed.

Notes

1. The form currently supports a single limit bucket and the first rate entry
2. More complex configurations (multiple limits, multiple rates, descriptors, etc.) can be handled in the YAML view
3. YAML-to-form sync works for supported fields and ignores unsupported nested structures

### DIFF
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -130,6 +130,14 @@
     }
   },
   {
+    "type": "console.page/route",
+    "properties": {
+      "exact": true,
+      "path": "/k8s/ns/:ns/ratelimitpolicies/:name/edit",
+      "component": { "$codeRef": "RateLimitPolicyEditPage" }
+    }
+  },
+  {
     "type": "console.resource/create",
     "properties": {
       "model": {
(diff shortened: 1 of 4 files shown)


R-Lawton:
Thank you for the interest, but this is work specifically for the LFX mentorship program, which is still in the process of picking mentees.  If you'd like to take part in the mentorship, please go through the right channels

PR #396, opened 2026-05-04
﻿fix: zero weight value silently dropped in DNSPolicy load balancing form

### LINKED ISSUE #395: fix: Zero weight value (0) silently dropped in DNSPolicy load balancing form
## Description

In the DNSPolicy creation form, entering `0` as the **Load Balancing Weight** value causes the field to be completely omitted from the generated YAML/spec with no warning or error shown to the user.

## Root Cause

In `src/components/KuadrantDNSPolicyCreatePage.tsx`, the weight is conditionally spread using a JavaScript **truthiness check**:

```ts
...(loadBalancing?.weight ? { weight: loadBalancing.weight } : {})
```

Since `0` is falsy in JavaScript, this condition evaluates to `{}` and the weight key is silently dropped from the policy spec entirely.

## Steps to Reproduce

1. Open the DNSPolicy creation page
2. Enable the **Load Balancing** section
3. Enter `0` in the **Weight** field
4. Switch to **YAML view**
5. Observe: the `weight` key is absent from the generated spec

## Expected Behavior

A weight of `0` is a valid Kubernetes value and should appear in the generated spec as:

```yaml
loadBalancing:
  weight: 0
```

## Actual Behavior

The `weight` key is missing entirely from the spec:

```yaml
loadBalancing: {}
```

## Proposed Fix

Replace the truthiness check with an explicit `null`/`undefined` guard:

```ts
// Before (buggy)
...(loadBalancing?.weight ? { weight: loadBalancing.weight } : {})

// After (correct)
...(loadBalancing?.weight !== null && loadBalancing?.weight !== undefined
  ? { weight: loadBalancing.weight }
  : {})
```

## Impact

Any user who intentionally sets a weight of `0` to deprioritize an endpoint will silently get incorrect behaviour the DNSPolicy will be created without the weight field, leading to unexpected load balancing results with no feedback in the UI.

## Environment

- **File:** `src/components/KuadrantDNSPolicyCreatePage.tsx`

befikre:
## What
Fixed a bug where a weight of 0 was silently dropped in the DNSPolicy creation form.

## Why
In JavaScript, 0 is falsy, so the existing truthiness check `loadBalancing?.weight ? ...` was evaluating to false and omitting the field from the generated YAML. This prevented users from deprioritizing endpoints using a 0 weight.

## Testing
- [x] Manually verified that entering 0 now appears correctly in the YAML view.
- [x] `yarn lint` passes.

Fixes #395

## Release Notes

* **Bug Fixes**
  * Fixed DNS Policy creation to properly include all weight configuration values in the generated output, even when set to zero.

### DIFF
--- a/src/components/KuadrantDNSPolicyCreatePage.tsx
+++ b/src/components/KuadrantDNSPolicyCreatePage.tsx
@@ -99,7 +99,9 @@ const KuadrantDNSPolicyCreatePage: React.FC = () => {
         ...(hasLoadBalancing
           ? {
               loadBalancing: {
-                ...(loadBalancing?.weight ? { weight: loadBalancing.weight } : {}),
+                ...(loadBalancing?.weight !== null && loadBalancing?.weight !== undefined
+                  ? { weight: loadBalancing.weight }
+                  : {}),
                 ...(loadBalancing?.geo ? { geo: loadBalancing.geo } : {}),
                 ...(loadBalancing.defaultGeo !== ''
                   ? { defaultGeo: loadBalancing.defaultGeo }


src/components/KuadrantDNSPolicyCreatePage.tsx
```diff
@@ -99,7 +99,9 @@ const KuadrantDNSPolicyCreatePage: React.FC = () => {
         ...(hasLoadBalancing
           ? {
               loadBalancing: {
-                ...(loadBalancing?.weight ? { weight: loadBalancing.weight } : {}),
+                ...(loadBalancing?.weight !== null && loadBalancing?.weight !== undefined
```
eguzki, line 102: `!= null `(loose equality) checks for both `null` and `undefined`

eguzki (changes requested):
This is a nice one! Nice catch. 

Specially because the default value of the weight is 120, so the user setting 0 in the form would end up having 120. https://github.com/Kuadrant/kuadrant-operator/blob/main/api/v1/dnspolicy_types.go#L101

eguzki:
superseded by https://github.com/Kuadrant/kuadrant-console-plugin/pull/515

PR #405, opened 2026-05-05
fix(config): replace script injection with JSON parsing to prevent XSS

### LINKED ISSUE #400: Security: Dynamic script injection in configLoader.ts creates XSS vulnerability
## Description
`configLoader.ts` fetches JavaScript from `config.js` and injects it 
directly into the DOM using `innerHTML`, which creates an XSS vulnerability.

## Location
`configLoader.ts` lines 28-30

## Current Code
```ts
const configScript = document.createElement('script');
configScript.innerHTML = script;  // dangerous
document.head.appendChild(configScript);
```

## Risk
If an attacker compromises the config endpoint or performs a 
MITM attack, they can execute arbitrary JavaScript in the OpenShift 
console context — potentially stealing auth tokens or accessing 
cluster resources.

## Suggested Fix
Replace executable JS config with a JSON-based approach:
```ts
const response = await fetch('/config.json');
const config = await response.json();
```
This eliminates script injection entirely.

## Environment
- Kuadrant Console Plugin (main branch)

anand-242003:
## Summary
Fixes the XSS vulnerability reported in #400.

Replaces the dynamic script injection pattern in `configLoader.ts` with 
secure JSON parsing. The previous approach fetched `config.js` and injected 
it directly into the DOM via `innerHTML`, allowing arbitrary code execution 
if the config endpoint was compromised.

## Changes
- `src/utils/configLoader.ts` — fetch `config.json`, parse with `response.json()`, validate structure before use
- `entrypoint.sh` — generate `config.json` instead of `config.js`
- `webpack.config.ts` — copy `config.json` to dist for local development
- `public/config.json` — default config for local development

## Breaking Change
Config endpoint changes from `/config.js` to `/config.json`.
`window.kuadrant_config` global is no longer set.

## Testing
1. Run locally with `yarn start`
2. Verify topology configmap loads correctly
3. Verify metrics suffix is applied correctly

Closes #400

## Summary by CodeRabbit

* **Refactor**
  * Improved configuration loading with enhanced error handling and fallback mechanisms
  * Configuration now validated at runtime with automatic defaults applied on failure
  * Configuration caching implemented for optimised performance

* **Chores**
  * Updated build process for streamlined asset management

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,12 +1,13 @@
 #!/bin/bash
 
-# Inject topology ConfigMap location and metrics configuration
-cat <<EOF > /tmp/config.js
-window.kuadrant_config = {
-  TOPOLOGY_CONFIGMAP_NAME: '${TOPOLOGY_CONFIGMAP_NAME:-topology}',
-  TOPOLOGY_CONFIGMAP_NAMESPACE: '${TOPOLOGY_CONFIGMAP_NAMESPACE:-kuadrant-system}',
-  METRICS_WORKLOAD_SUFFIX: '${METRICS_WORKLOAD_SUFFIX:-openshift-default}'
-};
+# Inject topology ConfigMap location and metrics configuration as JSON
+# Security: Using JSON instead of executable JavaScript to prevent XSS attacks
+cat <<EOF > /usr/share/nginx/html/config.json
+{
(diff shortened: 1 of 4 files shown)


R-Lawton:
is this actually solving an issue? For a MITM attack would they not have to have access to the clusters internal traffic?

PR #408, opened 2026-05-05
Standardize Kuadrant URL routes and namespace handling

### LINKED ISSUE #372: Unify Kuadrant URLs and Project selector
**Describe the bug**
At the moment, there are many pages under the developer portal section mainly that are not using the same namespace state with the OpenShift console Project Selector, due to URLs differences or internal state.

~~Here https://github.com/Kuadrant/kuadrant-console-plugin/blob/main/console-extensions.json#L325 should be `resource-ns`
and here https://github.com/Kuadrant/kuadrant-console-plugin/blob/main/src/components/KuadrantOverviewPage.tsx#L154-L188 we should remove the custom namespace selector~~

Independently of the _type_ of resource (k8s or aggregated with custom page) we need to unify the URLs under Kuadrant section

**Expected behavior**
- All pages should be consistent with the state set with the Project Selector injected by the OpenShift console or custom one.
- Kuadrant section should use _custom href_ and stick with them, not mixing _resource-ns_ and _custom href_.

didierofrivia:
Closes https://github.com/Kuadrant/kuadrant-console-plugin/issues/372

## Summary by CodeRabbit

* **New Features**
  - Namespace-aware navigation for Kuadrant Overview, Policies, Policy Topology and API Products with namespace selector and all-namespaces fallbacks.
  - API Products list links to related plan/policy items and supports namespaced create/edit flows and tabs.

* **Chores**
  - Console routes standardised to the new namespaced URL patterns.
  - End-to-end tests and UI navigation updated to follow the new routes and expectations.

### DIFF
--- a/src/components/KuadrantPoliciesPage.tsx
+++ b/src/components/KuadrantPoliciesPage.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { useState } from 'react';
-import { useNavigate, useLocation } from 'react-router-dom-v5-compat';
+import { useNavigate } from 'react-router-dom-v5-compat';
 import { useTranslation } from 'react-i18next';
 import { sortable } from '@patternfly/react-table';
 import {
@@ -341,31 +341,7 @@ const TLSPolicyTab: React.FC = () => {
 
 const KuadrantPoliciesPage: React.FC = () => {
   const { t } = useTranslation('plugin__kuadrant-console-plugin');
-  const location = useLocation();
-
-  // Extract current tab from URL to preserve it when changing namespaces
-  const getCurrentTab = React.useCallback(() => {
-    const pathParts = location.pathname.split('/').filter(Boolean);
-    const policiesIndex = pathParts.indexOf('policies');
-
-    if (policiesIndex !== -1 && pathParts.length > policiesIndex + 1) {
-      const nextSegment = pathParts[policiesIndex + 1];
-      // If next segment is not 'ns', it's a tab
-      if (nextSegment !== 'ns') {
-        return `/${nextSegment}`;
-      } else if (pathParts.length > policiesIndex + 3) {
-        // It's /kuadrant/policies/ns/namespace/tab pattern
-        return `/${pathParts[policiesIndex + 3]}`;
-      }
-    }
-
-    return '';
-  }, [location.pathname]);
-
-  const { handleNamespaceChange, activeNamespace } = useKuadrantNamespaceChange(
-    '/policies',
-    getCurrentTab,
-  );
+  const { handleNamespaceChange, activeNamespace } = useKuadrantNamespaceChange('/policies');
   const [activePerspective] = useActivePerspective();
 
   const defaultColumns: TableColumn<K8sResourceCommon>[] = [
--- a/src/components/KuadrantPoliciesPage.tsx
+++ b/src/components/KuadrantPoliciesPage.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { useState } from 'react';
-import { useParams, useNavigate } from 'react-router-dom-v5-compat';
+import { useNavigate, useLocation } from 'react-router-dom-v5-compat';
 import { useTranslation } from 'react-i18next';
 import { sortable } from '@patternfly/react-table';
 import {
@@ -8,7 +8,6 @@ import {
   HorizontalNav,
   TableColumn,
   K8sResourceCommon,
-  useActiveNamespace,
   useActivePerspective,
   ListPageCreateLink,
   ListPageBody,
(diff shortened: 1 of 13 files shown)


eguzki:
@coderabbitai review

src/components/KuadrantPoliciesPage.tsx
```diff
@@ -341,16 +341,32 @@ const TLSPolicyTab: React.FC = () => {
...
-    if (ns && ns !== activeNamespace) {
-      setActiveNamespace(ns);
+  const location = useLocation();
+
+  // Extract current tab from URL to preserve it when changing namespaces
+  const getCurrentTab = React.useCallback(() => {
+    const pathParts = location.pathname.split('/').filter(Boolean);
+    const policiesIndex = pathParts.indexOf('policies');
+
+    if (policiesIndex !== -1 && pathParts.length > policiesIndex + 1) {
```
eguzki: This looks complex. Wouldn't it be much simpler if we had different urls for all-namespaces and namespace scoped views? i.e.

```
/kuadrant/policies/ns/:ns
```

```
/kuadrant/all-namespaces/policies
```
This approach is consistent with OpenShift Console patterns

src/components/KuadrantPoliciesPage.tsx
```diff
@@ -341,16 +341,32 @@ const TLSPolicyTab: React.FC = () => {
...
-  const [activeNamespace, setActiveNamespace] = useActiveNamespace();
-  const [activePerspective] = useActivePerspective();
-  const navigate = useNavigate();
-
-  React.useEffect(() => {
-    if (ns && ns !== activeNamespace) {
-      setActiveNamespace(ns);
+  const location = useLocation();
+
+  // Extract current tab from URL to preserve it when changing namespaces
```
didierofrivia: The logic was similar in lines https://github.com/Kuadrant/kuadrant-console-plugin/pull/408/changes#diff-9639e7467201d4f0c99b905823f38b99d2f7852d2710aed22ba8e8f01c47ab20L465 We still might want to review this in the future.. see the example of other dynamic plugins

src/components/KuadrantPoliciesPage.tsx
didierofrivia: I could simplify it similar to how it was before, this is for tabs... what I commented here https://github.com/Kuadrant/kuadrant-console-plugin/pull/408/changes/BASE..4ac0fbc1c7fed08d0fafc7b0903d06a05c9e8bcd#r3197925517

> This approach is consistent with OpenShift Console patterns

it's consistent but it's not, that works if we are prefixing it with k8s/... following the example for k8s resources, but from our last call, we decided to scope everything in kuadrant/ ... then things get a bit more complicated.. specially with sub sections like tabs

eguzki: Sorry, I'm not following.

The prefix changes from `k8s` to `kuadrant`, but the URL pattern stays the same.

For example: deployments

* namespaced: `/k8s/ns/kuadrant-system/deployments`
* all namespaces: `/k8s/all-namespaces/deployments`

Same pattern for pretty much anything else in the console.

On the one hand, it's straightforward to read the active namespace from the URL (and update the NamespaceBar accordingly). On the other hand, when the user updates the namespace in the component, it's easy to navigate to that URL.

That would be for the list views. List views usually do not have tabs. Do they?

For resource details views (always namespaced), the OpenShift Console pattern is:
* Details view (main view) `/k8s/ns/kuadrant-system/deployments/my-deployment/`
* YAML tab view `/k8s/ns/kuadrant-system/deployments/my-deployment/yaml`

I suggest we follow the same pattern. If it works for them, it must work for us.

One interesting point you made that we must have as a feature: namespace is preserved between views. So if I am checking policies in namespace A and I click "api products" on the navigation menu, it should be preserved. OpenShift has a global state (active namespace) and we can rely on it. So when the URL says "all-namespaces" (that's the href of the navigation menu item) and the activeNamespace is set, honor the activeNamespace. Makes sense?

didierofrivia: that's what I suggested before this PR and we agreed on prefixing with _kuadrant_ ... there are some caveats on route and navigation that is kind of a handicap, they provide the prefix for href that is only _k8s_, not configurable and we are providing the navigation with the _namespaced_ prop, that adds the namespace at the end... take a look at the docs https://github.com/openshift/console/blob/main/frontend/packages/console-dynamic-plugin-sdk/docs/console-extensions.md#consolenavigationhref

R-Lawton:
👀

eguzki (commented):
Great work on the centralized routing hook!

LGTM

Left one nit, but feel free to ignore it.

It would be nice to have the url and href architecture documented. Specifically, how console.navigation/href with `namespaced: true` is being integrated with the useKuadrantNamespaceChange hook. And the custom URL pattern used (some of it already documented in the hook code). It deserves a markdown file. 

Locally tested, and active namespace is being persisted between views. 

When getting back to the policies page, having the tab view reseted is not an issue at all IMO. 

I see @R-Lawton is also reviewing, so I will leave final approval to her.

R-Lawton (dismissed):
verified it all works locally, the bug we were seeing now fixed. Nice job 🥇

R-Lawton (approved):

PR #434, opened 2026-05-10
fix: replace undefined 'history' with 'navigate' in useMemo deps for DNS/TLS policy action hooks

### LINKED ISSUE #433: fix: stale useMemo dependency - undefined 'history' referenced instead of 'navigate' in DNS/TLS policy action hooks
Describe the bug

In useDNSPolicyActions and useTLSPolicyActions, the useMemo dependency array references history which is undefined in scope. The hooks use useNavigate() (React Router v6) and store the result in a variable named navigate, but the dependency array was never updated from the old React Router v5 useHistory() API. Since history is always undefined, React treats it as a stable value and the memoized actions can hold a stale closure over the navigate function.

To Reproduce

1. Navigate to any DNSPolicy or TLSPolicy resource in the OpenShift Console
2. Open the actions menu (kebab menu) on a policy row
3. Click Edit (the form edit action)
4. Rapidly switch namespaces or trigger a route change while the actions menu is open
5. The navigate function inside the memoized action may be stale, causing incorrect navigation or a no-op
6. Expected behavior

The useMemo dependency array should include navigate so that whenever the navigate function reference changes, the memoized actions are recomputed with the latest function reference — consistent with the React hooks exhaustive-deps rule.

Additional context

- Affected files:
- src/components/dnspolicy/useDNSPolicyActions.tsx line 86
- src/components/tlspolicy/useTLSPolicyActions.tsx line 86

- Root cause: likely a leftover from a React Router v5 → v6 migration where useHistory() was replaced with useNavigate() and the variable renamed from history to navigate, but the useMemo deps array was not updated
- Fix: replace history with navigate in both dependency arrays

- }, [history, obj, dnsPolicyModel, launchAnnotationsModal, launchDeleteModal, launchLabelsModal]);
+ }, [navigate, obj, dnsPolicyModel, launchAnnotationsModal, launchDeleteModal, launchLabelsModal]);

shubhamkumar9199:
Fixes #433

The useMemo dependency arrays in useDNSPolicyActions and useTLSPolicyActions referenced history ,a variable that does not exist in scope. This is a leftover from the React Router v5 → v6 migration where useHistory() was replaced with useNavigate() and the variable renamed from history to navigate, but the useMemo dependency array was never updated.

Since history is always undefined, React treats it as permanently stable and never recomputes the memoized actions ,meaning the navigate function captured inside the memo can become stale, causing the Edit action on DNS/TLS policies to navigate incorrectly or silently fail.

Changes

- src/components/dnspolicy/useDNSPolicyActions.tsx - replace history with navigate in useMemo deps
- src/components/tlspolicy/useTLSPolicyActions.tsx - replace history with navigate in useMemo deps

- }, [history, obj, dnsPolicyModel, launchAnnotationsModal, launchDeleteModal, launchLabelsModal]);
+ }, [navigate, obj, dnsPolicyModel, launchAnnotationsModal, launchDeleteModal, launchLabelsModal]);
Test plan

- [X]  Open a DNSPolicy in the OpenShift Console and verify the Edit action navigates correctly to the form edit page
- [X]  Open a TLSPolicy and verify the same
- [X]  Verify no regression in Edit labels, Edit annotations, and Delete actions for both policy types

### DIFF
--- a/src/components/dnspolicy/useDNSPolicyActions.tsx
+++ b/src/components/dnspolicy/useDNSPolicyActions.tsx
@@ -83,7 +83,7 @@ const useDNSPolicyActions = (obj: K8sResourceCommon): ExtensionHookResult<Action
     ];
 
     return actionsList;
-  }, [history, obj, dnsPolicyModel, launchAnnotationsModal, launchDeleteModal, launchLabelsModal]);
+  }, [navigate, obj, dnsPolicyModel, launchAnnotationsModal, launchDeleteModal, launchLabelsModal]);
 
   return [actions, true, undefined];
 };
--- a/src/components/tlspolicy/useTLSPolicyActions.tsx
+++ b/src/components/tlspolicy/useTLSPolicyActions.tsx
@@ -83,7 +83,7 @@ const useTLSPolicyActions = (obj: K8sResourceCommon): ExtensionHookResult<Action
     ];
 
     return actionsList;
-  }, [history, obj, tlsPolicyModel, launchAnnotationsModal, launchDeleteModal, launchLabelsModal]);
+  }, [navigate, obj, tlsPolicyModel, launchAnnotationsModal, launchDeleteModal, launchLabelsModal]);
 
   return [actions, true, undefined];
 };


eguzki (approved):

eguzki:
Can you rebase @shubhamkumar9199  ? The required checks now include one build check that was recently added and it's not included in your branch.

shubhamkumar9199:
> Can you rebase @shubhamkumar9199 ? The required checks now include one build check that was recently added and it's not included in your branch.

@eguzki done

Issue #430, opened 2026-05-10
Hook Called Conditionally (Rules of Hooks violation)


shourya2006:
**Describe the bug**

In `KuadrantDNSPolicyCreatePage.tsx`, the `useK8sWatchResource` hook is called conditionally inside a ternary expression, which violates [React's Rules of Hooks](https://react.dev/reference/rules/rules-of-hooks#only-call-hooks-at-the-top-level). This causes a React invariant crash (`Rendered more hooks than during the previous render`) when a user navigates between the DNS Policy **edit** page and the DNS Policy **create** page within the same session, without doing a full page reload.

**Affected file:** `src/components/KuadrantDNSPolicyCreatePage.tsx` — lines 177–179

```typescript
// dnsResource is null when creating a new policy, non-null when editing
const [dnsData, dnsLoaded, dnsError] = dnsResource
  ? useK8sWatchResource(dnsResource)   // ← hook called conditionally {Which is wrong}
  : [null, false, null];
```

**To Reproduce**

> Requires Kuadrant CRDs to be installed on the cluster and at least one existing DNSPolicy resource.

1. Navigate to **Kuadrant → Policies → DNS** in the OpenShift console.
2. Click on an existing DNS Policy to open its detail page, then click **Edit** (or navigate directly to `/k8s/ns/<namespace>/kuadrant.io~v1~DNSPolicy/<policy-name>/yaml`). This loads the component with the hook **active** (edit mode).
3. Without refreshing the page, navigate back to **Kuadrant → Policies → DNS → Create DNS Policy** using the in-app navigation (sidebar or breadcrumb). This reuses the same component instance but now the hook is **skipped** (create mode).
4. React detects the hook count mismatch and throws a runtime error — the page crashes or goes blank.

**Expected behavior**

Navigating between the DNS Policy create page and edit page should work seamlessly without any crashes. The component should handle both create and edit modes without changing the number of React hooks called between renders.

**Screenshots**

The crash produces the following error in the browser DevTools console:

```
Uncaught Error: Rendered more hooks than during the previous render.
    at Object.throwInvalidHookError (react-dom.development.js)
```

The page renders blank or displays a React error boundary fallback.

**Additional context**

- This is a static code violation of [React's Rules of Hooks](https://react.dev/reference/rules/rules-of-hooks) — hooks must not be called conditionally, in loops, or inside nested functions.
[...]

PR #436, opened 2026-05-11
 Fix for Hook Called Conditionally (Rules of Hooks violation)

shourya2006:
Check https://github.com/Kuadrant/kuadrant-console-plugin/issues/430

1. Removed the conditional dnsResource variable + conditional hook
2. Added early return guard in useEffect
3. Added nameEdit to useEffect dependency array


shourya2006:
@R-Lawton check this out please

R-Lawton:
Thank you for contribution, but we could not reproduce issue

PR #439, opened 2026-05-11
fix: replace hardcoded path indexes with URL utility functions in DNS/TLS create pages

### LINKED ISSUE #440: fix: replace hardcoded URL path indexes with utility functions in DNSPolicy and TLSPolicy create pages
KuadrantDNSPolicyCreatePage and KuadrantTLSCreatePage use hardcoded array indexes (pathSplit[3], pathSplit[6]) to extract namespace and name from the URL. If the URL structure changes, these indexes silently return wrong values causing the edit form to load incorrect data.

To Reproduce

1. Go to a DNSPolicy or TLSPolicy resource in the OpenShift Console
2. Click "Edit" from the actions menu
3. The edit form loads - it relies on pathSplit[3] and pathSplit[6] to read the namespace and name from the URL
4. Expected behavior
5. The edit form should use the existing extractResourceNameFromURL and extractNamespaceFromURL utility functions which correctly locate URL segments using regex, regardless of URL format changes.

Additional context
The codebase already has these utility functions in src/utils/nameFromPath.ts. They were introduced to safely parse OpenShift Console URLs but were not applied to these two form pages. PR #408 standardised routes across the plugin - this is a related oversight in the same area.

shubhamkumar9199:
KuadrantDNSPolicyCreatePage and KuadrantTLSCreatePage were using hardcoded array indexes (pathSplit[3], pathSplit[6]) to extract namespace and name from the URL. This is fragile - if the URL structure changes (as it did in #408), these indexes silently return wrong values with no error. The codebase already has extractResourceNameFromURL and extractNamespaceFromURL utility functions that use regex to correctly locate these segments. This fix replaces the raw index access with those existing utilities.

Fixes #440

Type of change

- [X] - [fix] Bug fix

Changes made

- Replaced pathSplit[6] and pathSplit[3] with extractResourceNameFromURL and extractNamespaceFromURL in KuadrantDNSPolicyCreatePage.tsx
- Replaced pathSplit[6] and pathSplit[3] with extractResourceNameFromURL and extractNamespaceFromURL in KuadrantTLSCreatePage.tsx

Screenshots

- No visual change - this is an internal refactor to use existing URL utility functions.

Checklist

- [x]  All user-facing strings use t() and added to locales/en/plugin__kuadrant-console-plugin.json
- [x]  CSS classes are prefixed with kuadrant- (no global selectors)
- [x]  Commit(s) include Signed-off-by (git commit -s)

## Summary by CodeRabbit

* **Refactor**
  * Improved internal URL parsing logic in DNS Policy and TLS Policy creation pages for more consistent and maintainable handling of resource identifiers from URLs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-console-plugin/pull/439)

### DIFF
--- a/src/components/KuadrantDNSPolicyCreatePage.tsx
+++ b/src/components/KuadrantDNSPolicyCreatePage.tsx
@@ -33,6 +33,7 @@ import GatewaySelect from './gateway/GatewaySelect';
 import * as yaml from 'js-yaml';
 import KuadrantCreateUpdate from './KuadrantCreateUpdate';
 import { handleCancel } from '../utils/cancel';
+import extractResourceNameFromURL, { extractNamespaceFromURL } from '../utils/nameFromPath';
 import { resourceGVKMapping } from '../utils/resources';
 
 const KuadrantDNSPolicyCreatePage: React.FC = () => {
@@ -59,9 +60,8 @@ const KuadrantDNSPolicyCreatePage: React.FC = () => {
   const [creationTimestamp, setCreationTimestamp] = React.useState('');
   const [resourceVersion, setResourceVersion] = React.useState('');
   const location = useLocation();
-  const pathSplit = location.pathname.split('/');
-  const nameEdit = pathSplit[6];
-  const namespaceEdit = pathSplit[3];
+  const nameEdit = extractResourceNameFromURL(location.pathname);
+  const namespaceEdit = extractNamespaceFromURL(location.pathname);
   const [formDisabled, setFormDisabled] = React.useState(false);
   const [create, setCreate] = React.useState(true);
   const [loadBalancingExpanded, setLoadBalancingExpanded] = React.useState(false);
--- a/src/components/KuadrantTLSCreatePage.tsx
+++ b/src/components/KuadrantTLSCreatePage.tsx
@@ -22,6 +22,7 @@ import {
 } from '@openshift-console/dynamic-plugin-sdk';
 import './kuadrant.css';
 import { handleCancel } from '../utils/cancel';
+import extractResourceNameFromURL, { extractNamespaceFromURL } from '../utils/nameFromPath';
 import { useNavigate, useLocation } from 'react-router-dom-v5-compat';
 import * as yaml from 'js-yaml';
 import { useTranslation } from 'react-i18next';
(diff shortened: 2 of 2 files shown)


shiavm006:
hey @shubhamkumar9199 did u test it on the edit page? the route is /k8s/ns/X/dnspolicy/name/Y/edit, no group~v1~kind segment in there. so i think extractResourceNameFromURL would return null since its regex only matches gvk segments?

shiavm006:
if nameEdit ends up null, the if (nameEdit) at line 168 is skipped, watch never fires, and the form stays in create mode even when the user clicked edit. could be wrong tho. if the goal was just to drop the hardcoded indexes, maybe useParams() would fit better since the route already declares :ns and :name. 
i mean @eguzki   would know better

shubhamkumar9199:
Good catch @shiavm006, you're right that extractResourceNameFromURL would return null here because the edit route /k8s/ns/:ns/dnspolicy/name/:name/edit doesn't have a GVK segment. That utility was designed for console-native URLs like /k8s/ns/X/kuadrant.io~v1~DNSPolicy/Y/policies.

The current hardcoded pathSplit[6] actually works for this specific route, but the cleaner fix would be useParams() since the route already declares :ns and :name:


const { ns: namespaceEdit, name: nameEdit } = useParams<{ ns: string; name: string }>();
That avoids hardcoded indexes and is resilient to route changes. Happy to update the PR with this approach if that's preferred.

shiavm006:
i guess we should wait for maintainers before making any changes further let them review once cause discussing this would be better than implementing

R-Lawton:
@shiavm006 is right what you have breaks the code not fixes anything

shubhamkumar9199:
@R-Lawton Yeah fair, my bad, I didn't properly check the edit page before pushing.

You're right about the cause: the edit route /k8s/ns/:ns/dnspolicy/name/:name/edit has no gvk segment, and extractResourceNameFromURL only matches gvk-style URLs, so it just returns null here. That means nameEdit is null, the if (nameEdit) never runs, the watch doesn't fire, and it stays stuck in create mode. So yeah, this breaks edit rather than fixing anything.

useParams() is the better move like @shiavm006 said, since the route already gives us :ns and :name:


const { ns: namespaceEdit, name: nameEdit } = useParams<{ ns: string; name: string }>();
That drops the hardcoded pathSplit[6] (which was the whole point of the PR) and doesn't depend on a gvk segment at all.

I'll switch it over and actually test it on the edit page this time. If you'd rather just keep pathSplit[6] as-is and close this, that's fine too, just let me know.

Still on it, will push the fix shortly.

R-Lawton:
Lets stick with the current implemtation, it works and hasnt had any issues. thank you for looking into it

PR #452, opened 2026-05-12
Fix stale RBAC state in useAccessReviews hook

### LINKED ISSUE #451: useAccessReviews returns stale RBAC data when resource prop changes
# useAccessReviews returns stale RBAC data when resource prop changes

## Describe the bug

The `useAccessReviews` hook does not re-run RBAC access checks when the `resource` prop changes because the `useEffect` dependency array is empty. This causes stale permission data to be returned after resource updates.

Current implementation:

```ts
useEffect(() => {
    ...
  };

  checkRBAC();
}, []);

### PR DESCRIPTION
**Pull request by WildTrio**
## Description

This PR fixes stale RBAC permission data in the `useAccessReviews` hook located at:

```txt
src/utils/resourceRBAC.tsx
```

Previously, the hook executed RBAC access checks only once on component mount because the `useEffect` dependency array was empty:

```ts
useEffect(() => {
  checkRBAC();
}, []);
```

As a result, when the `resource` prop changed, the hook did not re-run access reviews and continued returning outdated RBAC permissions.

This PR updates the dependency array to include `resource`, ensuring RBAC checks are refreshed whenever the resource list changes.

---

## Changes Made

- Added `resource` to the `useEffect` dependency array
- Added `setLoading(true)` before re-fetching RBAC data
- Kept the implementation minimal and consistent with the existing code style

---

## Before

```ts
useEffect(() => {
    ...
  };

  checkRBAC();
}, []);
```

---

## After

```ts
useEffect(() => {
    setLoading(true);

    ...
  };

  checkRBAC();
}, [resource]);
```

---

## Impact

This fix ensures:
- RBAC permissions stay in sync with updated resources
- UI actions reflect the latest access state
- stale RBAC data is no longer returned after resource changes

---

## Testing

Tested by:
1. Rendering the hook with an initial resource
2. Updating the `resource` prop dynamically
3. Verifying that RBAC checks are triggered again
4. Confirming updated permissions are reflected in `userRBAC`

---

## Related Issue

Fixes #451

## Release Notes

* **Bug Fixes**
  * Fixed role-based access control (RBAC) checks to properly display loading state during permission verification.
  * Improved RBAC checks to update dynamically when resources change, ensuring permissions are accurately reflected in real time.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-console-plugin/pull/452)

### DIFF
--- a/src/utils/resourceRBAC.tsx
+++ b/src/utils/resourceRBAC.tsx
@@ -15,6 +15,8 @@ function useAccessReviews(resource: Resource[]) {
 
   useEffect(() => {
     const checkRBAC = async () => {
+      setLoading(true);
+
       const results = await Promise.all(
         resource.flatMap(({ group, kind, namespace }) =>
           verbs.map(async (verb) => {
@@ -24,20 +26,26 @@ function useAccessReviews(resource: Resource[]) {
               verb,
               ...(namespace ? { namespace } : {}),
             });
-            return { key: `${kind}-${verb}`, isAllowed: result.status?.allowed };
+
+            return {
+              key: `${kind}-${verb}`,
+              isAllowed: result.status?.allowed,
+            };
           }),
         ),
       );
+
       const RBACMap = results.reduce((acc, { key, isAllowed }) => {
         acc[key] = isAllowed;
         return acc;
       }, {} as Record<string, boolean>);
+
       setUserRBAC(RBACMap);
       setLoading(false);
     };
 
     checkRBAC();
-  }, []);
+  }, [resource]);
 
   return { userRBAC, loading };
 }


WildTrio (commented):
<sub>Reviewed in [CodeRabbit Review Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-console-plugin/pull/452)</sub>

jasonmadigan:
thanks @WildTrio

src/utils/resourceRBAC.tsx
```diff
@@ -24,22 +26,28 @@ function useAccessReviews(resource: Resource[]) {
...
+            };
           }),
         ),
       );
+
       const RBACMap = results.reduce((acc, { key, isAllowed }) => {
         acc[key] = isAllowed;
         return acc;
       }, {} as Record<string, boolean>);
+
```
jasonmadigan, line 42: all three callers (`DropdownWithKebab`, `KuadrantOverviewPage`, `ResourceList`) create the resource array inline each render — new reference every time, `useEffect` fires every render, `setState` triggers re-render, repeat forever.

fix: stabilise the dependency. simplest option is `JSON.stringify(resource)` as the dep inside the hook so callers don't need to change.

src/utils/resourceRBAC.tsx
```diff
@@ -15,6 +15,8 @@ function useAccessReviews(resource: Resource[]) {
 
   useEffect(() => {
     const checkRBAC = async () => {
```
jasonmadigan, line 17: needs a race condition guard — effect can re-fire while a previous `checkRBAC` is in flight. standard pattern:

```ts
let cancelled = false;
// ... await ...
if (!cancelled) { setUserRBAC(RBACMap); setLoading(false); }
return () => { cancelled = true; };
```

also needs try/catch — a failed `checkAccess` currently leaves `loading: true` permanently.

jasonmadigan (changes requested):
infinite loop — see line comments.

R-Lawton:
Hey @WildTrio Thanks for the pr but it hasn't been worked on in a while, if still relevant, please open back up and make the relevant ask for changes

PR #455, opened 2026-05-12
Optimize accessResources memoization in ResourceList

### LINKED ISSUE #453: ResourceList recreates accessResources on every render causing unnecessary RBAC checks
# ResourceList recreates accessResources on every render causing unnecessary RBAC checks

## Describe the bug

The `ResourceList` component recreates the `accessResources` array on every render before passing it to `useAccessReviews`.

Current implementation:

```ts
const accessResources = resources.map((r) => ({
  ...r,
  kind: getResourceNameFromKind(r.kind),
  namespace: resolvedNamespace,
}));

const { userRBAC, loading: rbacLoading } = useAccessReviews(accessResources);
```

Since `.map()` creates a new array reference on every render, `accessResources` changes even when the underlying `resources` data has not changed.

After adding `[resource]` as a dependency in `useAccessReviews`, this causes RBAC checks to re-run unnecessarily on every render.

---

## To Reproduce

1. Render the `ResourceList` component
2. Trigger any state update causing a re-render (pagination, filtering, etc.)
3. Observe that `useAccessReviews` is triggered again
4. RBAC checks are executed repeatedly even when `resources` are unchanged

---

## Expected behavior

RBAC checks should only re-run when the actual resource data changes.

The `accessResources` array should remain stable across renders unless its dependencies change.

---

## Suggested Fix

Memoize `accessResources` using `React.useMemo`.

Example:

```ts
const accessResources = React.useMemo(
  () =>
    resources.map((r) => ({
      ...r,
      kind: getResourceNameFromKind(r.kind),
      namespace: resolvedNamespace,
    })),
  [resources, resolvedNamespace],
);
```

---

## Impact

This issue can cause:
- unnecessary RBAC API requests
- repeated access reviews
- avoidable component re-renders
- performance degradation

---

## Additional context

File:

```txt
src/components/ResourceList.tsx
```

WildTrio:
## Description

This PR optimizes the `ResourceList` component by memoizing the `accessResources` array before passing it to `useAccessReviews`.

Previously, `accessResources` was recreated on every render because `.map()` generates a new array reference each time:

```ts
const accessResources = resources.map((r) => ({
  ...r,
  kind: getResourceNameFromKind(r.kind),
  namespace: resolvedNamespace,
}));
```

After the `useAccessReviews` hook was updated to depend on `[resource]`, this caused RBAC access reviews to re-run unnecessarily on every component render, even when the resource data itself had not changed.

This PR fixes the issue by wrapping `accessResources` with `React.useMemo`, ensuring the array reference remains stable unless its dependencies actually change.

---

## Changes Made

- Memoized `accessResources` using `React.useMemo`
- Added dependencies:
  - `resources`
  - `resolvedNamespace`
- Preserved existing functionality with minimal code changes

---

## Before

```ts
const accessResources = resources.map((r) => ({
  ...r,
  kind: getResourceNameFromKind(r.kind),
  namespace: resolvedNamespace,
}));
```

---

## After

```ts
const accessResources = React.useMemo(
  () =>
    resources.map((r) => ({
      ...r,
      kind: getResourceNameFromKind(r.kind),
      namespace: resolvedNamespace,
    })),
  [resources, resolvedNamespace],
);
```

---

## Impact

This change prevents:
- unnecessary RBAC API requests
- repeated access reviews on every render
- avoidable re-renders
- performance degradation caused by unstable array references

---

## Testing

Tested by:
1. Rendering `ResourceList`
2. Triggering component re-renders through filtering and pagination
3. Verifying that `useAccessReviews` no longer re-runs unnecessarily when resource data is unchanged

---

## Related Issue

Fixes #453

## Summary by CodeRabbit

* **Bug Fixes**
  * Access control checks now update reliably when resources change.
  * Improved handling of loading state during access control verification to avoid stale indicators.

* **Performance**
  * Optimised generation of resource access lists to improve responsiveness.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-console-plugin/pull/455)

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-console-plugin/pull/455)


WildTrio:
@coderabbitai review

src/utils/resourceRBAC.tsx
```diff
@@ -24,20 +26,26 @@ function useAccessReviews(resource: Resource[]) {
...
         return acc;
       }, {} as Record<string, boolean>);
+
       setUserRBAC(RBACMap);
       setLoading(false);
     };
 
     checkRBAC();
-  }, []);
+  }, [resource]);
```
jasonmadigan, line 65: `KuadrantOverviewPage` (line 186-191) still creates `rbacResources` inline via `.map()` each render — not touched by this PR. this will infinite loop the same way #452 did, just in a different component.

also, the `useMemo` approach in `ResourceList` depends on `resources` being referentially stable from the parent. if the parent creates the array inline (e.g. `KuadrantPoliciesPage` passes `filteredResources.map(...)`) the memo recomputes every render anyway. the `JSON.stringify` approach inside the hook would have been more robust — protects all callers regardless.

src/utils/resourceRBAC.tsx
```diff
@@ -15,6 +15,8 @@ function useAccessReviews(resource: Resource[]) {
 
   useEffect(() => {
     const checkRBAC = async () => {
```
jasonmadigan, line 19: still needs the two things flagged on #452:

1. race condition guard — if `resource` changes while `checkRBAC` is in flight, stale result overwrites the fresh one:
```ts
let cancelled = false;
// ... await ...
if (!cancelled) { setUserRBAC(RBACMap); setLoading(false); }
return () => { cancelled = true; };
```

2. try/catch — a rejected `checkAccess` leaves `loading: true` permanently with no recovery.

jasonmadigan (changes requested):
fixes two of three callers but misses `KuadrantOverviewPage`. also still missing the race condition guard and error handling from the #452 review.

WildTrio:
Will fix the issue soon...

WildTrio:
Updated the implementation based on the review feedback.

Changes made:

* added race condition cleanup guard to prevent stale async updates
* added error handling for failed `checkAccess` calls
* switched dependency handling to `JSON.stringify(resource)` to avoid unnecessary reruns caused by unstable array references
* reverted caller-side memoization changes and centralized the fix within `useAccessReviews`

This should address the issues raised around stale updates, loading state handling, and unstable dependency references.

R-Lawton:
@jasonmadigan are you happy with the changes that are made?

jasonmadigan:
changes fine, but there's a lint error

R-Lawton:
cool fix the lint error @WildTrio and this can be merged

R-Lawton:
PR Looks abandonded

PR #480, opened 2026-05-20
feat: Add pagination to MyAPIKeys page

### LINKED ISSUE #474: Add pagination to MyAPIKeys page
The MyAPIKeys page should have pagination similar to the APIProduct list page.

**Current state:** MyAPIKeys page lacks pagination
**Expected:** Apply the same pagination pattern used in the APIProduct list page

This will improve consistency and usability for users with many API keys.

justin212407:
## Summary

Implement pagination on the My API Keys list page to match the existing pattern used in the API Products list page. This improves UX for users with many API keys by displaying a limited set per page and allowing navigation between pages.

#### `MyAPIKeysPage.tsx`

- Added pagination state:
  - `currentPage`
  - `perPage` (default: 10 items per page)

- Implemented pagination logic:
  - Calculate `startIndex` and `endIndex` based on current page and items per page
  - Slice `filteredData` to show only the current page's items (`paginatedData`)
  - Reset to last valid page if filtering or data changes leave the current page empty

- Added pagination handlers:
  - `onSetPage` → updates current page number
  - `onPerPageSelect` → updates items per page and resets to page 1

- Rendered `Pagination` component at the bottom of the table with:
  - 5 items per page
  - 10 items per page
  - 20 items per page

- Updated `VirtualizedTable` to use `paginatedData` instead of `filteredData`

- Added `Pagination` import from `@patternfly/react-core`

## Pattern Consistency

This implementation follows the same pagination pattern used in `APIProductsListPage.tsx` (around lines 230–270), including:

- `useState`-based pagination state management
- Array slicing logic for current-page rendering
- Automatic page adjustment when filters reduce result count
- PatternFly `Pagination` component usage
- `.kuadrant-pagination-left` styling consistency
- Per-page options: `5`, `10`, and `20`

### Manual Testing

Verified:
- Pagination controls appear after creating multiple API keys
- Page navigation works correctly
- Per-page selection updates displayed rows properly
- Filtering integrates correctly with pagination state

## Related Issues

**Resolves:** #474

## Summary by CodeRabbit

* **New Features**
  * Client-side pagination added to the API Keys page with configurable items-per-page and page navigation.
  * Pagination automatically adjusts when filter results change and resets to the first page when filters are edited or removed.
  * The table view now displays paginated results while preserving the full filtered dataset; pagination controls appear only when results exist.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-console-plugin/pull/480?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


eguzki (commented):
Thanks for this 🙏 

Consider doing something about:
- Consistent page reset on filter changes (prevents empty views)
- Memoized pagination calculation (minor performance optimization)
- Dual pagination (top toolbar + bottom) for better UX on long lists

There is has a potential bug where changing filters might leave the user on a page beyond the available data, showing an empty view until they manually paginate back.

Check APIProductAPIKeysTab.tsx for similar implementation

justin212407:
Sure will do that. I'll follow the same implementation as is done in `APIProductAPIKeysTab.tsx`.

eguzki:
It's not a request for changes. It's also a request for discussion. Your implementation might have aspects to be adopted by the other impl

justin212407:
I can update `MyAPIKeysPage` to align more closely with the `APIProductsListPage` behavior by resetting to page 1 on filter changes while still keeping the clamp safeguard for invalid pages.

One thing I did find useful in the current `MyAPIKeysPage` implementation is the “stay as close as possible to the current position” behavior through page clamping instead of always aggressively resetting pagination state. That approach may still be valuable in views where users are browsing large datasets and making smaller refinements rather than applying entirely new filters.

I also think the clamp-based safeguard itself is something that could be consistently applied across the pagination implementations, since `APIKeyApprovalTable` currently seems more vulnerable to invalid-page empty states if the backing data shrinks dynamically.

Since you mentioned there may be aspects of this implementation worth adopting elsewhere, I think the current approach helped highlight some useful pagination tradeoffs around page preservation vs reset behavior that could help shape a more consistent shared pattern across the list views.
@eguzki What are your views regarding this?

eguzki:
Good catch on the clamping safeguard - that **is** valuable and should be added to `APIKeyApprovalTable` and `APIProductAPIKeysTab` to prevent invalid-page states when data shrinks dynamically.

However, I'd push back on the "page preservation" framing. Not resetting to page 1 on filter changes isn't a deliberate UX pattern - it's missing functionality that happens to have a defensive safeguard. 

1. **Filtered data is a new dataset** - page 5 of unfiltered results has no semantic relationship to page 5 of filtered results
2. **Users expect to see filtered results from the beginning** - if I filter 100 items down to 15, I want to see items 1-15, not have the view clamp to page 2 showing items 11-15

**Recommendation:**
- Add clamping consistently to  `APIKeyApprovalTable` and other components using pagination.
- Add `setCurrentPage(1)` to MyAPIKeysPage filter handlers
- Don't adopt "page preservation on filter" as a pattern elsewhere

justin212407:
Framing it as “page preservation” was probably the wrong interpretation on my side. I agree that once filters change, the resulting dataset should be treated as a new result set, so resetting to page 1 is the more intuitive and consistent UX.

I’ll update `MyAPIKeysPage` to reset pagination on filter changes while keeping the clamping safeguard for dynamic data shrink scenarios.

I also agree the clamping logic would be valuable to standardize across the other pagination implementations to avoid invalid-page empty states.

justin212407:
> Good catch on the clamping safeguard - that **is** valuable and should be added to `APIKeyApprovalTable` and `APIProductAPIKeysTab` to prevent invalid-page states when data shrinks dynamically.
> 
> However, I'd push back on the "page preservation" framing. Not resetting to page 1 on filter changes isn't a deliberate UX pattern - it's missing functionality that happens to have a defensive safeguard.
> 
> 1. **Filtered data is a new dataset** - page 5 of unfiltered results has no semantic relationship to page 5 of filtered results
> 2. **Users expect to see filtered results from the beginning** - if I filter 100 items down to 15, I want to see items 1-15, not have the view clamp to page 2 showing items 11-15
> 
> **Recommendation:**
> 
> * Add clamping consistently to  `APIKeyApprovalTable` and other components using pagination.
> * Add `setCurrentPage(1)` to MyAPIKeysPage filter handlers
> * Don't adopt "page preservation on filter" as a pattern elsewhere

Updated `MyAPIKeysPage` to reset pagination to page 1 whenever filters/search criteria change, while retaining the existing clamping safeguard for invalid-page scenarios caused by dynamic data shrink.

This now aligns the pagination behavior more closely with `APIProductsListPage`:

* filter changes are treated as a new dataset and start from page 1
* clamping still protects against invalid pagination state during live data updates

While reviewing the pagination implementations for consistency, I also noticed that `APIKeyApprovalTable` currently doesn’t clamp invalid pages if the backing dataset shrinks dynamically. I kept that out of this PR to avoid expanding the scope, but I can open a small follow-up issue/PR to add the same safeguard there if that would be useful.

eguzki:
> While reviewing the pagination implementations for consistency, I also noticed that APIKeyApprovalTable currently doesn’t clamp invalid pages if the backing dataset shrinks dynamically. I kept that out of this PR to avoid expanding the scope, but I can open a small follow-up issue/PR to add the same safeguard there if that would be useful.

Sounds like a good idea

justin212407:
> > While reviewing the pagination implementations for consistency, I also noticed that APIKeyApprovalTable currently doesn’t clamp invalid pages if the backing dataset shrinks dynamically. I kept that out of this PR to avoid expanding the scope, but I can open a small follow-up issue/PR to add the same safeguard there if that would be useful.
> 
> Sounds like a good idea

On it. will create the issue and link the pr at the same place.

eguzki (approved):
LGTM

Good job on this one

justin212407:
Thanks

PR #489, opened 2026-05-25
fix(e2e): add extended timeout for kuadrant readiness in CI

R-Lawton:
## Summary
- oinc has a hardcoded 5-minute timeout for the Kuadrant CR `Ready` condition, which sometimes isn't enough on CI runners
- this causes flaky `e2e-rbac` failures during infrastructure setup — before any tests even run (e.g. [PR #488](https://github.com/Kuadrant/kuadrant-console-plugin/pull/488))
- catches the oinc failure, checks if the Kuadrant CR was already created, and waits up to 10 additional minutes for readiness
- on persistent failure, dumps pod status and CR yaml for easier debugging

## Test plan
- [ ] CI `e2e-rbac` job passes on this PR
- [ ] no behaviour change when kuadrant becomes ready within oinc's original 5-minute window

### DIFF
--- a/scripts/cluster-setup.sh
+++ b/scripts/cluster-setup.sh
@@ -25,7 +25,34 @@ PLUGIN_NAME=$(node -p "require('${REPO_DIR}/package.json').consolePlugin.name")
 log "creating oinc cluster with addons..."
 oinc create \
 	--addons gateway-api,cert-manager,metallb,istio,kuadrant@latest \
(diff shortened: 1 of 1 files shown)


eguzki:
I'm not sure it's a timing issue. I think it's more likely a bug in the kuadrant-operator. There's a race condition where sometimes the kuadrant-operator fails to detect the CRD dependencies and never marks the kuadrant CR as ready.

PR #504, opened 2026-05-29
Use multi-stage build to fix ppc64le/s390x image builds

thomasmaas:
## Summary

Fixes the ppc64le build failure in [release-build run 26587098293](https://github.com/Kuadrant/kuadrant-console-plugin/actions/runs/26587098293) caused by a known QEMU/OpenSSL incompatibility when running `dnf` under emulation.

## Root cause

The single-stage Dockerfile ran `dnf update && dnf install` under QEMU user-space emulation for non-amd64 architectures. On ppc64le, OpenSSL's TLS handshake fails under QEMU with:

```
Curl error (35): SSL connect error for https://cdn-ubi.redhat.com/...
[error:030000EA:digital envelope routines::provider signature failure]
```

This is a [widely reported issue](https://forums.docker.com/t/qemu-breaks-with-openssl-fips/148270) affecting QEMU-emulated containers using OpenSSL on UBI9.

## Fix

Split the Dockerfile into two stages:

1. **Build stage** (`--platform=linux/amd64`): Runs Node.js, yarn, and the JS build natively on amd64. The output (static HTML/JS/CSS) is architecture-independent — no need to emulate.

2. **Runtime stage** (target architecture): Uses `ubi9-minimal` with only nginx installed via `microdnf`. No `dnf`, no Node.js toolchain. Much smaller image.

## Additional benefits

- **Smaller runtime image**: `ubi-minimal` + nginx only, instead of full UBI9 + Node.js + gcc + python3
- **Removed `dnf update/upgrade`**: These pulled in unnecessary packages and slowed every build
- **Cleaner separation**: Build tools don't ship in the final image

## Test plan
- [ ] amd64 image builds and runs correctly
- [ ] arm64 image builds and runs correctly
- [ ] ppc64le image builds successfully (was failing before)
- [ ] s390x image builds successfully
- [ ] Plugin loads in OpenShift console and serves static assets
- [ ] `entrypoint.sh` generates `/tmp/config.js` correctly
- [ ] Nginx serves from `/usr/share/nginx/html/` as expected

## Summary by CodeRabbit

* **Chores**
  * Optimised Docker build process with explicit multi-stage architecture targeting and streamlined dependency installation.
  * Added build-time verification checks for critical distribution artefacts.
  * Restructured runtime container image for improved efficiency and reduced footprint.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-console-plugin/pull/504?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,17 @@
-FROM registry.access.redhat.com/ubi9/ubi:latest
+# Stage 1: Build static assets on native amd64 (avoids QEMU emulation issues
+# with OpenSSL/TLS when running dnf on ppc64le/s390x under emulation)
+FROM --platform=linux/amd64 registry.access.redhat.com/ubi9/ubi:latest AS builder
 
 USER root
 
-RUN dnf update -y && \
-    dnf upgrade -y && \
-    dnf module enable nodejs:22 nginx:1.24 -y && \
-    dnf install -y nodejs nginx npm gcc-c++ make python3
+RUN dnf module enable nodejs:22 -y && \
+    dnf install -y nodejs npm gcc-c++ make python3 && \
+    dnf clean all
 
-# Enable corepack for Yarn v4 support
 RUN npm install -g corepack && \
     corepack enable
 
 RUN yarn config set --home enableGlobalCache true
-
-RUN mkdir -p /var/cache/nginx /var/log/nginx /run && \
-    chmod -R 777 /var/cache/nginx /var/log/nginx /run /usr/share/nginx/html/
 
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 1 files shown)


eguzki (commented):
This is going in a good direction having two building stages, so the runtime stage has a smaller image. 

I am approving this but I have a concern on this new image running on openshift. Before merging, I would like to see this running on openshift. When you deploy a container to OpenShift, it doesn't necessarily run as the UID you specify in the Dockerfile (USER 1001). Instead, OpenShift's Security Context Constraints (SCCs) typically assign a random UID at runtime.

R-Lawton:
@eguzki did you get to see this run in openshift and are happy to merge ?

eguzki (approved):
LGTM

It's working in openshift environments.

The image has right permissions to run and the nginx paths have correct permissions for nginx to serve content. Also the entrypoint.sh script correctly creates the config.js that is being served as configuration for the console app.

PR #523, opened 2026-06-03
ci: pin kuadrant e2e addon version and dump diagnostics on setup failure

namansh70747:
## Description

The `e2e-rbac` workflow intermittently fails during cluster setup with `Error: addon installation: kuadrant not ready: kuadrant not ready after 5m0s` (#361). The CI logs show `deployment ready` for `kuadrant-operator-controller-manager` and then time out, with no indication of *what* `oinc` is still waiting on — so each occurrence is opaque and unactionable.

The readiness wait itself lives inside the [`oinc`](https://github.com/jasonmadigan/oinc) tool (pinned to `v0.1.2` in [e2e.yaml](.github/workflows/e2e.yaml)), not in this repo, so this PR does **not** change the timeout. Instead it makes the failure reproducible and debuggable from this side.

Refs #361

## Type of change

- [x] `[chore]` Dependency / config update (CI / e2e setup)

## Changes made

- Pin the kuadrant e2e addon to a fixed version (`1.4.4`) instead of `@latest`. `@latest` silently drifts as new operator releases land (it resolved to `1.4.1` at the time of the linked failure), which makes attributing a flaky-readiness failure to a specific version impossible. Override with `KUADRANT_VERSION=latest` to test against the newest.
- On `oinc create` failure, dump kuadrant addon diagnostics — the `Kuadrant` CR conditions, pod status (`get`/`describe`), and operator logs. `oinc` waits on more than the operator `Deployment`, so the existing "deployment ready → timeout" logs don't reveal the actual blocker.

## Test plan

- [x] `bash -n scripts/cluster-setup.sh` passes (syntax check)
- [x] Change is confined to `scripts/cluster-setup.sh`; no app/UI/i18n code touched
- [ ] Full e2e run on CI (to confirm the diagnostic path and pinned version)

## Notes

This is a diagnostic + reproducibility improvement, not a root-cause fix — the next failure will carry CR conditions and operator logs so we can see whether it's a webhook registration delay, a `developerPortal`-component readiness gap, or runner capacity. Happy to also bump `OINC_VERSION` to `v0.1.3` in a follow-up, though its changelog only covers a console-plugin hostname fix, not readiness.

## Release Notes

* **Chores**
  * Allow Kuadrant addon version to be pinned via an environment variable (defaults to latest) for reproducible cluster setups.
  * Improved cluster-setup failure diagnostics: automatically gather addon state, pod status, namespace events and operator logs when deployments fail to aid troubleshooting.

### DIFF
--- a/scripts/cluster-setup.sh
+++ b/scripts/cluster-setup.sh
@@ -12,6 +12,11 @@ source "${SCRIPT_DIR}/lib.sh"
 
 PLUGIN_PORT="${PLUGIN_PORT:-9001}"
 
+# the console plugin tracks the latest Kuadrant CRDs, so default to the latest
+# operator. override (e.g. KUADRANT_VERSION=1.4.4) to pin to a known-good version
+# if a latest release breaks plugin development or CI (GH-361).
+KUADRANT_VERSION="${KUADRANT_VERSION:-latest}"
+
 check_command oinc "Install from https://github.com/jasonmadigan/oinc"
 check_command kubectl "Install from https://kubernetes.io/docs/tasks/tools/"
 
@@ -22,10 +27,28 @@ HOST=$(container_host "${RUNTIME}")
 
 PLUGIN_NAME=$(node -p "require('${REPO_DIR}/package.json').consolePlugin.name")
 
-log "creating oinc cluster with addons..."
-oinc create \
-	--addons gateway-api,cert-manager,metallb,istio,kuadrant@latest \
-	--console-plugin "${PLUGIN_NAME}=http://${HOST}:${PLUGIN_PORT}"
+# on a failed cluster create, dump kuadrant addon state so the "kuadrant not
+# ready after 5m0s" timeout (GH-361) is debuggable from CI logs instead of
+# opaque. oinc waits on more than the operator deployment, so capture the
+# Kuadrant CR conditions, namespace events, and the operator logs.
+dump_kuadrant_diagnostics() {
+	log "oinc create failed - dumping kuadrant addon diagnostics for GH-361..."
+	kubectl get kuadrant kuadrant -n kuadrant-system -o yaml 2>&1 || true
+	kubectl get pods -n kuadrant-system -o wide 2>&1 || true
+	kubectl get events -n kuadrant-system --sort-by='.lastTimestamp' 2>&1 || true
+	kubectl logs deployment/kuadrant-operator-controller-manager -n kuadrant-system --tail=200 --all-containers 2>&1 || true
+}
+
+# bash suspends `set -e` (errexit) for a command used as an `if` condition, so a
+# failed `oinc create` won't abort here - it falls through to the diagnostics
+# dump and an explicit exit instead of dying silently.
+log "creating oinc cluster with addons (kuadrant@${KUADRANT_VERSION})..."
+if ! oinc create \
+	--addons "gateway-api,cert-manager,metallb,istio,kuadrant@${KUADRANT_VERSION}" \
+	--console-plugin "${PLUGIN_NAME}=http://${HOST}:${PLUGIN_PORT}"; then
+	dump_kuadrant_diagnostics
+	exit 1
+fi
 
 log "patch kuadrant to enable developer portal controller..."
 kubectl patch kuadrant kuadrant -n kuadrant-system --type merge --patch '{"spec": {"components": {"developerPortal": {"enabled": true}}}}'


scripts/cluster-setup.sh
```diff
@@ -22,10 +27,25 @@ HOST=$(container_host "${RUNTIME}")
...
+dump_kuadrant_diagnostics() {
+	log "oinc create failed — dumping kuadrant addon diagnostics for GH-361..."
+	kubectl get kuadrant -n kuadrant-system -o yaml 2>&1 || true
+	kubectl get pods -n kuadrant-system -o wide 2>&1 || true
+	kubectl describe pods -n kuadrant-system 2>&1 || true
+	kubectl logs -n kuadrant-system -l app=kuadrant --tail=200 --all-containers 2>&1 || true
+}
+
+log "creating oinc cluster with addons (kuadrant@${KUADRANT_VERSION})..."
+if ! oinc create \
```
eguzki, line 46: add a comment explaining how this work.

When a command appears in a conditional context (like if, while, &&, ||), bash temporarily suspends the -e (errexit) behavior for that specific command.

scripts/cluster-setup.sh
```diff
@@ -22,10 +27,25 @@ HOST=$(container_host "${RUNTIME}")
...
+# on a failed cluster create, dump kuadrant addon state so the "kuadrant not
+# ready after 5m0s" timeout (GH-361) is debuggable from CI logs instead of
+# opaque. oinc waits on more than the operator deployment, so capture the
+# Kuadrant CR conditions and the operator pods/logs.
+dump_kuadrant_diagnostics() {
+	log "oinc create failed — dumping kuadrant addon diagnostics for GH-361..."
+	kubectl get kuadrant -n kuadrant-system -o yaml 2>&1 || true
+	kubectl get pods -n kuadrant-system -o wide 2>&1 || true
+	kubectl describe pods -n kuadrant-system 2>&1 || true
+	kubectl logs -n kuadrant-system -l app=kuadrant --tail=200 --all-containers 2>&1 || true
```
eguzki: Maybe better would be read kuadrant operator logs directly

```
kubectl logs -f deployment/kuadrant-operator-controller-manager -n kuadrant-system
```

scripts/cluster-setup.sh
```diff
@@ -12,6 +12,11 @@ source "${SCRIPT_DIR}/lib.sh"
 
 PLUGIN_PORT="${PLUGIN_PORT:-9001}"
 
+# pin the kuadrant addon version so runs are reproducible. `@latest` drifts as
```
eguzki: IMHO this comment adds noise rather than give context to a future reader.

scripts/cluster-setup.sh
```diff
@@ -12,6 +12,11 @@ source "${SCRIPT_DIR}/lib.sh"
...
+# new operator releases land, which makes flaky-readiness failures (GH-361) hard
+# to attribute. override with KUADRANT_VERSION=latest to test against the newest.
+KUADRANT_VERSION="${KUADRANT_VERSION:-1.4.4}"
```
eguzki: The Kuadrant console plugin relies on CRDs provided by the Kuadrant operator. So, for daily use, I am in favor of using "latest" so the latest version of CRDs is available. Version 1.4.4 of the Kuadrant operator does not have the CRDs this console plugin requires in the HEAD of `main`. Very soon, v1.5.0 will be released and this repo could pin to v1.5.0 for the required CRDs. In the future, if we update or add new CRDs, `latest` would again be needed. Regarding the mentioned stability, that's a good point. If something breaks in the latest version of the Kuadrant operator, it would block the development of the console plugin. This is why I like having the KUADRANT_VERSION configurable in case it's needed. If the latest version of Kuadrant is broken, we can always pin to a working version to unblock the console plugin's development. For the stability of the CI checks, I agree with your thinking. We should pin to a working version. That has to be v1.5.0. That can be done in the e2e/setup.sh script that calls scripts/cluster-setup.sh. Unfortunately, since it's just a few days away from release, we cannot pin Kuadrant to v1.5.0 today for running the CI e2e workflow.

eguzki (commented):
Added some comments

scripts/cluster-setup.sh
namansh70747: Good point — added a comment above the `if` explaining that bash suspends `set -e` (errexit) for a command in a conditional context, so a failed `oinc create` falls through to the diagnostics dump and an explicit `exit 1` rather than aborting silently.

scripts/cluster-setup.sh
namansh70747: Done — switched to `kubectl logs deployment/kuadrant-operator-controller-manager -n kuadrant-system --tail=200 --all-containers` (dropped the `-f` so it doesn't hang the script in CI). This also resolves the bots' concern that the previous `-l app=kuadrant` selector might not match any pods.

scripts/cluster-setup.sh
namansh70747: Removed the original comment and replaced it with a shorter one that just explains the default + override rationale per your CRD point below.

scripts/cluster-setup.sh
namansh70747: Thanks for the detailed context — that makes sense, and I wasn't aware `1.4.4` lacked the CRDs `main` needs. Reverted the default back to `latest` so daily dev and CI track the latest CRDs, and kept `KUADRANT_VERSION` configurable purely as an escape hatch to pin a known-good version if a `latest` release breaks things. Agreed that CI pinning to v1.5.0 belongs in `e2e/setup.sh` once it's released — happy to follow up with that PR when v1.5.0 lands.

scripts/cluster-setup.sh
```diff
@@ -22,10 +27,25 @@ HOST=$(container_host "${RUNTIME}")
...
-oinc create \
-	--addons gateway-api,cert-manager,metallb,istio,kuadrant@latest \
-	--console-plugin "${PLUGIN_NAME}=http://${HOST}:${PLUGIN_PORT}"
+# on a failed cluster create, dump kuadrant addon state so the "kuadrant not
+# ready after 5m0s" timeout (GH-361) is debuggable from CI logs instead of
+# opaque. oinc waits on more than the operator deployment, so capture the
+# Kuadrant CR conditions and the operator pods/logs.
+dump_kuadrant_diagnostics() {
+	log "oinc create failed — dumping kuadrant addon diagnostics for GH-361..."
+	kubectl get kuadrant -n kuadrant-system -o yaml 2>&1 || true
```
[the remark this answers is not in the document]
namansh70747: Fixed — now targets the specific CR with `kubectl get kuadrant kuadrant -n kuadrant-system -o yaml`.

scripts/cluster-setup.sh
```diff
@@ -22,10 +27,25 @@ HOST=$(container_host "${RUNTIME}")
...
-log "creating oinc cluster with addons..."
-oinc create \
-	--addons gateway-api,cert-manager,metallb,istio,kuadrant@latest \
-	--console-plugin "${PLUGIN_NAME}=http://${HOST}:${PLUGIN_PORT}"
+# on a failed cluster create, dump kuadrant addon state so the "kuadrant not
+# ready after 5m0s" timeout (GH-361) is debuggable from CI logs instead of
+# opaque. oinc waits on more than the operator deployment, so capture the
+# Kuadrant CR conditions and the operator pods/logs.
+dump_kuadrant_diagnostics() {
+	log "oinc create failed — dumping kuadrant addon diagnostics for GH-361..."
```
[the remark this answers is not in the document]
namansh70747: Fixed — replaced the em dash with an ASCII hyphen in the log line.

scripts/cluster-setup.sh
```diff
@@ -22,10 +27,25 @@ HOST=$(container_host "${RUNTIME}")
...
+	kubectl get kuadrant -n kuadrant-system -o yaml 2>&1 || true
+	kubectl get pods -n kuadrant-system -o wide 2>&1 || true
+	kubectl describe pods -n kuadrant-system 2>&1 || true
```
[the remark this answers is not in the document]
namansh70747: Agreed — dropped the full-namespace `describe pods`. The diagnostics now rely on `get pods -o wide`, `get events --sort-by=.lastTimestamp`, and the operator deployment logs, which are more targeted and less noisy.

scripts/cluster-setup.sh
```diff
@@ -22,10 +27,25 @@ HOST=$(container_host "${RUNTIME}")
...
+	kubectl logs -n kuadrant-system -l app=kuadrant --tail=200 --all-containers 2>&1 || true
+}
```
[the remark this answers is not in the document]
namansh70747: Addressed — added `kubectl get events -n kuadrant-system --sort-by=.lastTimestamp` and replaced the `-l app=kuadrant` log selector (which may not match) with the operator deployment directly: `kubectl logs deployment/kuadrant-operator-controller-manager`.

namansh70747:
Thanks for the thorough reviews @eguzki, plus Copilot and CodeRabbit. Pushed a revision addressing everything:

**Version handling** (per @eguzki's CRD point)
- Reverted the default back to `kuadrant@latest` so daily dev and CI track the latest CRDs that `main` needs.
- Kept `KUADRANT_VERSION` configurable as an escape hatch to pin a known-good version if a `latest` release breaks development or CI.
- Replaced the "noisy" comment with a shorter one explaining just the default + override rationale.
- CI pinning to v1.5.0 in `e2e/setup.sh` — agreed, happy to follow up once v1.5.0 is released.

**Diagnostics**
- `get kuadrant kuadrant` (specific CR, not all) — Copilot.
- Dropped full-namespace `describe pods`; added `get events --sort-by=.lastTimestamp` — Copilot / CodeRabbit.
- Read operator logs directly via `deployment/kuadrant-operator-controller-manager` instead of the `-l app=kuadrant` selector (which may not match) — @eguzki / CodeRabbit.
- ASCII hyphen instead of em dash — Copilot.

**Readability**
- Added a comment explaining that bash suspends `set -e` for a command in an `if` condition — @eguzki.

Also rebased onto the latest `main`.

eguzki (approved):

PR #541, opened 2026-06-05
fix: garbage-collect API key Secret by setting an ownerReference

### LINKED ISSUE #512: API Key Secret is orphaned on delete — no ownerReference and no rollback on partial failure
The API Key request flow creates a Secret (holding a live credential) and an APIKey as two independent objects. The Secret has no ownerReference to the APIKey, and deletion removes only the APIKey so the credential Secret is orphaned. A partial failure during creation leaves the same orphan. (Name-collision handling is tracked separately in [#404](https://github.com/Kuadrant/kuadrant-console-plugin/issues/404) / [#481](https://github.com/Kuadrant/kuadrant-console-plugin/issues/481) and is out of scope here.)

shiavm006:
## Description
- Create the `APIKey` first, then create the `Secret` with an `ownerReference` to it, so Kubernetes garbage-collects the Secret automatically when the APIKey is deleted.
- Roll back the APIKey if Secret creation fails, so a partial failure doesn't leave a key without its Secret.

Reordering is safe: the APIKey references the Secret by name (`secretRef.name`), which is known up front.

Fixes #512

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API key request flow with better error recovery and automatic resource cleanup. The system now properly manages the lifecycle of API keys, ensuring associated resources are automatically removed when an API key is deleted.

### DIFF
--- a/src/components/apikey/RequestAPIKeyModal.tsx
+++ b/src/components/apikey/RequestAPIKeyModal.tsx
@@ -27,6 +27,7 @@ import {
   useK8sWatchResource,
   useActiveNamespace,
   k8sCreate,
+  k8sDelete,
   useK8sModel,
 } from '@openshift-console/dynamic-plugin-sdk';
 import { RESOURCES, APIKey, Secret } from '../../utils/resources';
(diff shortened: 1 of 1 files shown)


eguzki:
The code looks good. However, I found an issue while testing that I reported in https://github.com/Kuadrant/developer-portal-controller/issues/78

The issue is not in the code of this PR, but the changes of this PR activated the issue. We need to address https://github.com/Kuadrant/developer-portal-controller/issues/78 before we merge this issue or the api creation workflow would be broken. 

I also tested the edge case of pre-existing secret. The apikey is being deleted and the issue is notified to the user. 
Screenshot 2026-07-01 at 15-00-39 My API Keys

I think it is correct to catch `AlreadyExists` and report as error. Before these errors were omitted.

eguzki:
https://github.com/Kuadrant/developer-portal-controller/issues/78 has been fixed.

eguzki (approved):
LGTM

PR #558, opened 2026-06-21
fix: do not emit empty loadBalancing block on every DNSPolicy

shubhamkumar9199:
hasLoadBalancing was computed from `loadBalancing.weight != null`, but weight defaults to 120, so the condition was always true. As a result every DNSPolicy created via the form serialized a `spec.loadBalancing` block (with just `weight: 120` and no geo) even when the user never opened the Load Balancing section.

Gate on `loadBalancing.geo` instead, which is the field the form validation already requires for load balancing and which is populated both when the user fills the section and when an existing policy is loaded for editing.
 
## Type of change

- [x] `[fix]` Bug fix
- [ ] `[feat]` New feature
- [ ] `[refactor]` Refactor (no functional changes)
- [ ] `[test]` Test updates
- [ ] `[chore]` Dependency / config update

## Test plan

- [ ] `yarn lint` passes (no changes after running)
- [ ] `yarn build` passes
- [ ] `yarn i18n` passes (no changes after running — commit updated locale files if needed)
- [ ] Tested manually in OpenShift Console
- [ ] Tested in both light and dark themes
- [ ] Tested in all-namespaces and single namespace mode

## Checklist

## Summary

* **Bug Fixes**
  * Improved DNS Policy creation so incomplete load balancing settings are no longer saved when the geographical distribution isn’t configured.
  * Load balancing defaults are now applied consistently when creating or reloading a policy, and the load balancing section is only generated when the geo option is set to avoid partial/invalid configurations.

### DIFF
--- a/src/components/KuadrantDNSPolicyCreatePage.tsx
+++ b/src/components/KuadrantDNSPolicyCreatePage.tsx
@@ -208,7 +208,7 @@ const KuadrantDNSPolicyCreatePage: React.FC = () => {
         setProviderRefs([providerRef]);
         setLoadBalancing({
           geo: dnsPolicyUpdate.spec?.loadBalancing?.geo || '',
-          weight: dnsPolicyUpdate.spec?.loadBalancing?.weight ?? null,
+          weight: dnsPolicyUpdate.spec?.loadBalancing?.weight ?? 120,
           defaultGeo:
             dnsPolicyUpdate.spec?.loadBalancing?.defaultGeo !== undefined
               ? dnsPolicyUpdate.spec.loadBalancing?.defaultGeo
@@ -244,7 +244,7 @@ const KuadrantDNSPolicyCreatePage: React.FC = () => {
 
       setLoadBalancing({
         geo: parsedYaml.spec?.loadBalancing?.geo || '',
-        weight: parsedYaml.spec?.loadBalancing?.weight ?? null,
+        weight: parsedYaml.spec?.loadBalancing?.weight ?? 120,
         defaultGeo:
           parsedYaml.spec?.loadBalancing?.defaultGeo !== undefined
             ? parsedYaml.spec.loadBalancing?.defaultGeo
--- a/src/components/KuadrantDNSPolicyCreatePage.tsx
+++ b/src/components/KuadrantDNSPolicyCreatePage.tsx
@@ -35,6 +35,9 @@ import KuadrantCreateUpdate from './KuadrantCreateUpdate';
 import { handleCancel } from '../utils/cancel';
 import { resourceGVKMapping } from '../utils/resources';
 
+// Default weight applied to weighted load-balancing endpoints (matches the CRD default).
+const DEFAULT_LOAD_BALANCING_WEIGHT = 120;
+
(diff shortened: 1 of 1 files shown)


emmaaroche (commented):
Please create an issue for PRs being opened if there isn't one already.

Found what I feel is a UX issue during testing where the weight now shows as placeholder only and the user must manually type `120` for it to serialize.

Placeholder text can be confusing as it looks filled in but isn't actually a value. Better approach might be to keep weight pre-filled at `120` (actual value, not placeholder), but only serialize the `loadBalancing` block when `geo` is filled. This way:
- Weight 120 is visible as the sensible default
- No confusion about whether field is filled or not
- Geo is required for load balancing anyway, so gating serialization on geo makes sense

So the implementation would be to initialize `weight` state to `120` (actual value, not placeholder), keep the `hasLoadBalancing` gate on `geo` (as this PR does), and when geo is filled serialize the entire block including weight. WDYT?

shubhamkumar9199:
> Please create an issue for PRs being opened if there isn't one already.
> 
> Found what I feel is a UX issue during testing where the weight now shows as placeholder only and the user must manually type `120` for it to serialize.
> 
> Placeholder text can be confusing as it looks filled in but isn't actually a value. Better approach might be to keep weight pre-filled at `120` (actual value, not placeholder), but only serialize the `loadBalancing` block when `geo` is filled. This way:
> 
> * Weight 120 is visible as the sensible default
> * No confusion about whether field is filled or not
> * Geo is required for load balancing anyway, so gating serialization on geo makes sense
> 
> So the implementation would be to initialize `weight` state to `120` (actual value, not placeholder), keep the `hasLoadBalancing` gate on `geo` (as this PR does), and when geo is filled serialize the entire block including weight. WDYT?

Thanks for testing this. You're right about the placeholder confusion, and I tracked down the cause. Once serialization is gated on geo, a fresh form produces YAML with no loadBalancing block, so syncing the form back from that YAML (or edit-loading a policy with no stored weight) ran weight through ?? null, which wiped the 120 default down to just the placeholder

src/components/KuadrantDNSPolicyCreatePage.tsx
```diff
@@ -206,7 +208,7 @@ const KuadrantDNSPolicyCreatePage: React.FC = () => {
         setProviderRefs([providerRef]);
         setLoadBalancing({
           geo: dnsPolicyUpdate.spec?.loadBalancing?.geo || '',
-          weight: dnsPolicyUpdate.spec?.loadBalancing?.weight ?? null,
+          weight: dnsPolicyUpdate.spec?.loadBalancing?.weight ?? 120,
```
emmaaroche: The magic number 120 appears 4 times throughout this file. Maybe we could extract it to a named constant at the top of the file? Something like: `const DEFAULT_LOAD_BALANCING_WEIGHT = 120;`

shubhamkumar9199: Pulled the default into `const DEFAULT_LOAD_BALANCING_WEIGHT = 120` at the top of the file and pointed the initial state and both fallbacks at it, so the default now lives in one place.

emmaaroche (approved):
LGTM 👍

PR #598, opened 2026-07-06
test: add unit tests for API key, metrics, and topology utilities

### LINKED ISSUE #566: Add unit tests for API key, metrics, and topology utilities
## Summary

Write unit tests for the pure utility functions that handle API key status, PromQL query building, and topology graph parsing. All functions in scope are pure — no mocking required.

## Testing Guidance

For each function, think about:
- What does it return for normal, expected input?
- What happens at the boundaries — empty strings, missing segments, undefined values?
- If there's branching logic (if/else, switch), have you covered each branch?

Name your tests so someone reading the output knows exactly what broke, e.g. `"returns null when limits is undefined"` not `"test formatLimits"`.

### `src/utils/apiKeyUtils.ts`
`formatLimits` — formats rate limits into a readable string. There's a priority order to which limit type gets displayed — look at the if/else chain. Consider what happens with undefined, empty objects, and zero values. Pay attention to the `!== undefined` check — it behaves differently to a truthy check when the value is `0`. The `custom` branch has different structure to the named limit branches.

### `src/components/apikey/utils.ts`
`getRequestStatus` — similar pattern to `getAPIKeyPhase` but with a 3-state model. Check the priority order and the fallback.

### `src/utils/metricsQueries.ts`
Three query builder functions (`buildTotalRequestsQuery`, `buildErrorRequestQuery`, `buildErrorsByCodeQuery`). The key branching is with/without a namespace argument. These produce PromQL query strings where syntax correctness matters — mismatched braces or wrong regex patterns break the metrics dashboard silently (empty data, not an error). Test that the output queries are valid PromQL structure.

### `src/utils/topology/graphParser.ts`
- `kindToAbbr` — extracts uppercase letters with a regex, falls back to `toUpperCase()` for all-lowercase input, and truncates to 4 characters. Three chained operations — test the fallback path and truncation behavior.
- `parseDotToModel` — parses DOT strings into topology models. Node labels follow a `"Kind\\nName"` format. Think about what happens with unconnected nodes (grouping logic) and edges referencing non-existent nodes.
- `preserveTransitiveEdges` — bridges connections when intermediate nodes are filtered out, but only for specific node types. Check `transitiveNodeTypes` to see which ones. Think about what happens when no nodes are filtered, when transitive nodes are filtered, and when non-transitive nodes are filtered.

## Review & Verification

PRs should include:
- How to review the changes (areas of interest, suggested review order)
- Steps to confirm locally: `yarn test` runs and passes

R-Lawton:
## Summary

- adds 50 unit tests across 4 co-located test files for pure utility functions
- adds a single jest ESM stub for SDK and PatternFly topology (no behavioural mocking, just so jest can parse the imports)
- jest config updated with moduleNameMapper entries for the stub

Fixes https://github.com/Kuadrant/kuadrant-console-plugin/issues/566

## How to review

suggested order:
1. `__mocks__/esmStubs.ts` — single stub file, minimal exports to stop ESM parse failures
2. `jest.config.ts` — two moduleNameMapper additions pointing to the same stub
3. `src/utils/apiKeyUtils.test.ts` — formatLimits, start simple
4. `src/components/apikey/utils.test.ts` — getRequestStatus
5. `src/utils/metricsQueries.test.ts` — PromQL query builders + buildGatewayKey
6. `src/utils/topology/graphParser.test.ts` — most complex, topology parsing + transitive edges

## Test plan

- [ ] `yarn test` runs and all 50 tests pass

## Summary by CodeRabbit

* **Tests**
  * Added coverage for API key status and limit formatting logic.
  * Added query-builder tests to verify namespace filtering and key generation.
  * Added topology parsing tests for node grouping, edge handling, and invalid input cases.

* **Bug Fixes**
  * Improved test support for ESM-only dependencies so the suite runs more reliably.
  * Added stubs for external topology and console-related packages used during testing.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -14,6 +14,9 @@ const config: Config = {
   moduleNameMapper: {
     // CSS and static file imports
     '\\.(css|svg|png|jpg|jpeg|gif)$': '<rootDir>/__mocks__/fileMock.ts',
+    // ESM modules that Jest cannot parse — stub with no-op exports
+    '^@openshift-console/dynamic-plugin-sdk$': '<rootDir>/__mocks__/esmStubs.ts',
+    '^@patternfly/react-topology$': '<rootDir>/__mocks__/esmStubs.ts',
   },
   testMatch: ['**/*.test.{ts,tsx}'],
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx'],
--- a/__mocks__/esmStubs.ts
+++ b/__mocks__/esmStubs.ts
@@ -1,0 +1,20 @@
+// Stub for Jest — these modules use ESM exports that Jest cannot parse.
+// Tests that need specific behaviour should add their own jest.mock() overrides.
+
+// @openshift-console/dynamic-plugin-sdk
+export const k8sCreate = () => Promise.resolve({});
+export const k8sUpdate = () => Promise.resolve({});
+export const k8sList = () => Promise.resolve([]);
+export const k8sDelete = () => Promise.resolve({});
+export const k8sGet = () => Promise.resolve({});
+export const consoleFetchJSON = Object.assign(() => Promise.resolve({}), {
+  post: () => Promise.resolve({}),
+});
+export const useK8sModel = () => [null, false];
+export const useK8sWatchResource = () => [[], true, null];
+
+// @patternfly/react-topology
+export const NodeShape = { rect: 'rect', ellipse: 'ellipse' };
+export const LabelPosition = { bottom: 'bottom' };
+export const EdgeStyle = { default: 'default', dashedMd: 'dashedMd' };
+export const EdgeAnimationSpeed = { medium: 'medium' };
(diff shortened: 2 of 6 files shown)


jest.config.ts
```diff
@@ -14,6 +14,9 @@ const config: Config = {
   moduleNameMapper: {
     // CSS and static file imports
     '\\.(css|svg|png|jpg|jpeg|gif)$': '<rootDir>/__mocks__/fileMock.ts',
+    // ESM modules that Jest cannot parse — stub with no-op exports
+    '@openshift-console/dynamic-plugin-sdk': '<rootDir>/__mocks__/esmStubs.ts',
```
jasonmadigan: these are unanchored regexes, so `@openshift-console/dynamic-plugin-sdk` also matches `@openshift-console/dynamic-plugin-sdk-webpack` (a real dep) and any subpath imports, which would get silently routed to the stub. probably worth anchoring for safety.

jasonmadigan (dismissed):
a small nit, but this looks good

jasonmadigan (approved):

PR #601, opened 2026-07-06
fix: remove invalid targetRef.namespace and rename endpoint to path

### LINKED ISSUE #594: Policy create forms emit invalid spec.targetRef.namespace (DNS/TLS/OIDC)
## Summary

The DNSPolicy, TLSPolicy, and OIDCPolicy create/edit forms serialize a `namespace` field inside `spec.targetRef`. That field does not exist on the Kuadrant policy `targetRef` type. It is silently pruned by the API server in the default case, hard-rejected under strict field validation, and shown to users in the live YAML preview as if it were valid.

## Affected code

- `src/components/KuadrantDNSPolicyCreatePage.tsx` (`namespace: selectedGateway.namespace` in the targetRef builder)
- `src/components/KuadrantTLSCreatePage.tsx`
- `src/components/KuadrantOIDCPolicyCreatePage.tsx`
- Plus the matching `namespace?: string` added to each file's local edit interface, and the edit-load / YAML-parse fallbacks that read `spec.targetRef.namespace`.

Introduced by d6698a4 (DNS/TLS) and aecc093 (OIDC).

## Why it is a bug

1. **CRD type.** `DNSPolicy.spec.targetRef` is a Gateway API `LocalPolicyTargetReferenceWithSectionName`: fields are `group`, `kind`, `name`, `sectionName`, with no `namespace`, and it can only refer to objects in the same namespace as the policy. TLSPolicy and OIDCPolicy use the same local reference type.
2. **Internal consistency.** The other four policy create pages (Auth, RateLimit, TokenRateLimit, Plan) emit `targetRef: { group, kind, name }` with no namespace. Only the three recently-touched pages differ.
3. **Fixtures.** No policy `targetRef` in `e2e/manifests/test-resources.yaml` declares a namespace.
4. **History.** Commit 27ef64e ("fix: revert namespace changes") already removed this field from the OIDC page during review; aecc093 later re-added it.

## Impact

- Default clusters: structural-schema pruning drops the unknown field server-side, so the create succeeds but the field never persists.
- Strict field validation (or applying the same YAML via GitOps): the create is rejected with "unknown field spec.targetRef.namespace".
- The YAML editor shows an invalid resource, including `namespace: ''` before a gateway is selected (the value is emitted unconditionally, unlike the conditionally-spread healthCheck/loadBalancing blocks).
- On edit or Form/YAML round-trip, the fallback re-injects the invalid field into policies that did not have it.

## Underlying requirement to preserve

`GatewaySelect` lists gateways cluster-wide and keys options as `namespace/name`, and the FormSelect value is `selectedGateway.namespace/selectedGateway.name`. d6698a4 added `targetRef.namespace` so the gateway selection survived a Form/YAML toggle. Any fix must keep that persistence without writing an invalid field. Because the target must be same-namespace, the gateway namespace always equals the policy namespace, so it is derivable from `metadata.namespace` and never needs to live in `targetRef`.

## Proposed fix

Recommended (option A, matches the earlier 27ef64e precedent, lowest-risk):
1. Remove `namespace` from `spec.targetRef` in the three create pages and drop `namespace?` from their edit interfaces.
2. On edit-load and YAML-parse, set `selectedGateway.namespace` from the policy namespace (`metadata.namespace` / active namespace) instead of `targetRef.namespace`, restoring toggle persistence the correct way.

Optional follow-up (option D): scope `GatewaySelect` to the policy namespace so users cannot select a cross-namespace gateway that would produce an invalid policy.

## Acceptance criteria

- Creating a DNS/TLS/OIDC policy from the form produces a `spec.targetRef` with only group/kind/name (and optional sectionName), matching the other policy forms and the e2e fixtures.
- The YAML preview never shows `targetRef.namespace` (including no `namespace: ''`).
- Gateway selection still survives toggling between Form and YAML views.

emmaaroche:
## Summary

Fixes DNS/TLS/OIDC policy forms emitting invalid CRD fields rejected by admission webhooks.

## Changes

**Removed `spec.targetRef.namespace` from DNS/TLS/OIDC policies**

The Gateway API `LocalPolicyTargetReferenceWithSectionName` type only supports `group`, `kind`, `name`, and `sectionName` fields. The `namespace` field doesn't exist and gets rejected under field validation.

The target gateway must be in the same namespace as the policy per the local reference spec, so we derive the gateway namespace from `metadata.namespace` instead.

**Renamed DNSPolicy `healthCheck.endpoint` to `healthCheck.path`**

The DNSPolicy CRD spec defines the field as `path`, not `endpoint`. Forms were emitting the wrong field name, causing webhook rejection. Verified against CRD - the field has always been `path` for health check path configuration.

image

## Testing

Created DNS/TLS/OIDC policies via form - no webhook errors, valid YAML output.

Fixes #594

## Summary by CodeRabbit

* **New Features**
  * Health check settings now use a **Path** field instead of **Endpoint**.
  * New policy forms no longer include a namespace on the selected gateway reference.

* **Bug Fixes**
  * Editing and YAML import now consistently use the resource’s namespace when loading existing policy data.
  * Health check values are now read and saved using the updated path field throughout the create/edit flow.


Anton-Fil:
👀

jasonmadigan (approved):

PR #605, opened 2026-07-07
migrate yarn scripts to makefile targets

jasonmadigan:
yarn v4 drops support for custom scripts in package.json, breaking
workflows for contributors on newer yarn versions. this wraps all
yarn scripts as make targets so the build interface works regardless
of yarn version.

colons in yarn script names (e.g. `test:e2e`) mapped to hyphens
(`test-e2e`) since colons are reserved in make. existing `oinc` and
`oinc-teardown` targets preserved. `make help` lists all targets.


R-Lawton:
🤔 why close?

PR #622, opened 2026-07-09
test: replace waitForLoadState with element wait in smoke tests

emmaaroche:
## Description

Three `@smoke` tests in `apiproduct-crud.spec.ts` were using `waitForLoadState('networkidle')` after SPA navigate (`pushState + popstate`). This is the same root cause as #621 — `networkidle` resolves when there's no network activity, but React Router rendering happens synchronously so the component may not have mounted yet. These tests were observed failing in CI under parallel execution for the same reason.

Replaced with `expect(page.locator('#display-name')).toBeVisible()` which waits for the actual form input to appear, confirming the page has fully rendered before interacting. Also updates the same pattern in `should create APIProduct via form and verify in list` (from #621) to use the modern locator-based API consistently.

Also adds `@smoke` tag to `should sync between Form and YAML views` and fixes an ambiguous locator (`text=API key approval` matched both a nav link and a form heading — narrowed to `getByRole('heading')`).

After tagging `should sync between Form and YAML views` as `@smoke` it was caught failing in CI with two distinct errors:

1. **`#display-name` not found (20s timeout)** — the local `spaNavigate` function in `apiproduct-crud.spec.ts` was a duplicate of the one in `helpers.ts` but was missing the `waitForKuadrantPlugin` call. This meant SPA navigation could fire before the Kuadrant plugin had registered its routes, so the create form never mounted.

2. **Modal backdrop blocking tags button click (retry)** — the OpenShift console tour can re-trigger on the first visit to a plugin page. `dismissConsoleTour` was only called in `beforeEach` at `/`. On retry, the form loaded but the tour modal appeared mid-test and blocked subsequent clicks.

   `should sync between Form and YAML views` is the only test that performs a second `page.goto` in the test body (to set `activeNamespace` before SPA navigation). This full page reload resets the console's "first visit" state and can re-trigger the tour. Other smoke tests only have the `beforeEach` goto to `/`, so one `dismissConsoleTour` is sufficient and SPA navigation does not re-trigger the modal.

Fixed by importing `spaNavigate` from `helpers.ts` (which includes the Kuadrant plugin wait) instead of the local duplicate, and adding `dismissConsoleTour` after navigation in the yaml-sync test.

## Changes
- Replace `waitForLoadState('networkidle')` with `expect(locator).toBeVisible()` in 4 smoke tests
- Add `@smoke` tag to `should sync between Form and YAML views`
- Fix ambiguous `text=API key approval` locator to `getByRole('heading', { name: 'API Key approval' })`
- Modernize all `page.waitForSelector` calls to locator-based API
- Replace local `spaNavigate` (missing Kuadrant plugin wait) with the exported version from `helpers.ts`
- Import `navigateToAPIProducts` from `helpers.ts` instead of keeping a local duplicate
- Add `dismissConsoleTour` after navigation in `should sync between Form and YAML views` to handle late-appearing tour modals

## Testing
1. Run `npx playwright test --config=e2e/playwright.config.ts e2e/tests/apiproduct-crud.spec.ts`
2. Verify all 12 pass

### DIFF
--- a/e2e/tests/apiproduct-crud.spec.ts
+++ b/e2e/tests/apiproduct-crud.spec.ts
@@ -241,12 +241,12 @@ test.describe('APIProduct CRUD Operations', () => {
     }
   });
 
-  test('should sync between Form and YAML views', async ({ page }) => {
+  test('should sync between Form and YAML views', { tag: '@smoke' }, async ({ page }) => {
     // Full page load first so activeNamespace is set correctly before SPA navigation
     await page.goto(`/k8s/ns/${TEST_NAMESPACE}`);
     await page.waitForLoadState('networkidle');
     await navigateToAPIProductCreate(page, TEST_NAMESPACE);
-    await page.waitForLoadState('networkidle');
+    await page.waitForSelector('#display-name', { state: 'visible', timeout: 20000 });
 
     // Fill form fields
     await page.locator('#display-name').fill('YAML Sync Test');
@@ -579,7 +579,7 @@ EOF`, { stdio: 'inherit' });
 
   test('should display validation messages for required fields', { tag: '@smoke' }, async ({ page }) => {
     await navigateToAPIProductCreate(page, TEST_NAMESPACE);
-    await page.waitForLoadState('networkidle');
+    await page.waitForSelector('#display-name', { state: 'visible', timeout: 20000 });
 
     // Leave display name empty and try to proceed
     const saveButton = page.locator('button:has-text("Create")');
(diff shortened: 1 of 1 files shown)


Anton-Fil:
👀

emmaaroche:
Tests still acting flaky in CI - investigating more

Anton-Fil:
The test failure is not caused by the changes in this PR. This is a pre-existing flaky failure tracked in #626 — should delete APIProduct with confirmation fails intermittently in CI because the newly created
 APIProduct doesn't appear in the list table within the timeout, either due to the OpenShift Console Watch/Informer delay or pagination (resource lands on a page other than the first). Emma's changes did not 
 touch this logic.

PR #630, opened 2026-07-10
fix(e2e): fix apiproduct-crud intermittent failures

### LINKED ISSUE #626: e2e: Night test failed
**Describe the bug**
`should delete APIProduct with confirmation` fails intermittently in CI. The test creates an APIProduct via `kubectl apply` in `beforeEach`, then navigates to the API Products list and waits for the resource row to
appear — but the row is never found within the timeout and the test fails.

**To Reproduce**
Run the full e2e suite with `workers: 3`. The delete test fails with:

Error: expect(locator).toBeVisible() failed
Locator: locator('tr:has-text("test-edit-...")')
Timeout: 10000ms

**Expected behavior**
The newly created APIProduct should appear in the list and the delete flow should complete successfully.

**Additional context**
`e2e/tests/apiproduct-crud.spec.ts`
Test: `should delete APIProduct with confirmation` (nested inside `edit and delete` describe)
link in git hab workflou for fail: https://github.com/Kuadrant/kuadrant-console-plugin/actions/runs/29066373268

Possible causes:
- The OpenShift Console Watch/Informer stream may not reflect a `kubectl apply`-created resource within the 10-second timeout
- Pagination — the resource may exist but land on a page other than  the first one (the `create` test had to add full pagination retry  logic for exactly this reason)
- The `edit` test above it avoids this entirely by navigating directly to the resource URL rather than searching the list

emmaaroche:
## Description

`should delete APIProduct with confirmation` was failing intermittently in CI. The resource row was looked up with a single `toBeVisible` call, which fails if the resource lands on a pagination page other than the first - which happens when other tests have already created products. The create test already had pagination retry logic; this PR extracts it into a `findRowWithPagination` helper and applies it to the delete test too. As a side effect, the retry loop (up to 10 attempts with 1s waits) also gives the OpenShift watch stream time to deliver the `kubectl apply` event before giving up.

## Changes

- Add `findRowWithPagination` helper that iterates through all pagination pages, waiting between attempts for the watch stream to deliver the event
- Use it in both the delete test (replacing the broken `toBeVisible` timeout) and the create test (deduplicating the existing inline loop)

Closes #626

## Testing

- All 12 `apiproduct-crud.spec.ts` tests pass locally with `workers: 3`

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end coverage for API Product creation and deletion across paginated product lists.
  * Added more reliable validation that products can be found regardless of which page they appear on.
  * Enhanced deletion workflow checks by consistently locating the relevant product before confirmation steps.

### DIFF
--- a/e2e/tests/apiproduct-crud.spec.ts
+++ b/e2e/tests/apiproduct-crud.spec.ts
@@ -7,6 +7,27 @@ async function navigateToAPIProductCreate(page: Page, namespace = 'kuadrant-test
   await dismissConsoleTour(page);
 }
 
+// Scroll through all pagination pages looking for a row matching `text`.
+// Retries up to `maxAttempts` times, waiting for the watch stream between pages.
+async function findRowWithPagination(page: Page, text: string, maxAttempts = 10): Promise<boolean> {
+  const row = page.locator(`tr:has-text("${text}")`);
+  for (let attempt = 0; attempt < maxAttempts; attempt++) {
+    if (await row.isVisible()) return true;
+    const nextBtn = page.locator('button[aria-label="Go to next page"]');
+    if ((await nextBtn.count()) > 0 && !(await nextBtn.isDisabled())) {
+      await nextBtn.click();
+      await page.waitForTimeout(500);
+    } else {
+      await page.waitForTimeout(1000);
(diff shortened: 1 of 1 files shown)


R-Lawton:
👀

R-Lawton (approved):

Issue #670, opened 2026-07-17
MCP management overview page


R-Lawton:
Implement the populated MCP management overview page as shown in the Figma designs. The empty state stub is added in Kuadrant/kuadrant-console-plugin#666; this issue builds it out with data.

**Sections (top to bottom):**

1. **Page header** — `Title`: "MCP management overview"

2. **Get started banner**
   - `Alert` (info variant) with "Get started with Kuadrant:" text and external documentation link
   - Kebab menu (`Dropdown`) with "Don't show again" option

3. **Summary cards**
   - `Card` split into two sections: MCP Gateways | MCP Servers
   - MCP Gateways: Total, Healthy (green check), Unhealthy (warning triangle)
   - MCP Servers: Types count, Total, Online (green check), Offline (warning triangle)

4. **MCP gateway traffic** ⚠️ blocked on mcp-gateway observability
   - `Card` with `Select` (gateway filter: "All gateways") + `Button` (time range) + tooltip icon
   - Line chart (requests over time, multi-series per gateway with legend)
   - Stat tiles: Requests/min, Success rate, Error rate, Avg latency — each with % change from last day

5. **MCP Gateway Extensions table**
   - `Toolbar`: `Select` filter (Name) + `SearchInput` + pagination
   - `Table` columns: Extension name (link), Gateway name (with tooltip icon), Namespace (`Label`), Reference Grant (link)
   - Kebab actions per row: Edit, Delete (confirmation `Modal` → `k8sDelete`)
   - `Button` (primary, split): "Create extension"

6. **MCP Servers table**
   - `Toolbar`: `Select` (Name, with settings icon) + `SearchInput` + pagination
   - `Table` columns: Server name (link), Namespace (`Label`), Status (`Label` with icon), Details (badges showing tool count, resource count, prompt count, and/or action links like "Create policy", "Policy management")
   - Kebab actions per row: Edit, Delete (confirmation `Modal` → `k8sDelete`)
   - `Button` (primary, split): "Register MCP Server"

7. **Reference grants table**
   - `Toolbar`: `Select` (Name, with settings icon) + `SearchInput` + pagination
   - `Table` columns: Grant name (link), Namespace mapped to (`Label`)
   - Kebab actions per row: Edit, Delete (confirmation `Modal` → `k8sDelete`)
   - `Button` (primary): "Create reference grant"

8. **Policies attached to MCP gateways or servers table**
[...]

Issue #667, opened 2026-07-17
Epic: MCP management UI designs


R-Lawton:
Right now managing MCP Gateways means writing YAML and running oc commands. There's no way to see what's registered, test tools, or register new servers without switching between the terminal, YAML files, and external tooling like MCP Inspector.

This epic adds an MCP management section to the OpenShift console — browse servers, inspect tools and prompts, test tool execution, register new servers through a guided form, and see gateway health at a glance. One place to operate the gateway instead of three.

**Designs:** https://www.figma.com/proto/oizjCN6EVed31dW71YXS3K/RHCL-MCP?node-id=137-138949

## Issues

- #666 Setup wizard: MCP infrastructure setup. Done.
- #673 Setup wizard: MCP server registration. Done.
- #670 MCP management overview page. Done.
- #668 Standalone resource pages: MCPGatewayExtension, MCPServerRegistration and HTTPRoute. Done.
- #669 Standalone resource pages: Gateway and Reference grant. Done.
- #753 Validate the direct-CORS PoC. Closed, replaced by #776.
- #776 Validate the MCP Inspector backend proxy approach. Done.
- #752 MCP Inspector client and gateway discovery foundation. Superseded by #776 and #779.
- #671 MCP Inspector: Tools. In progress in #779 and Kuadrant/kuadrant-operator#2206.
- #672 MCP Inspector: Prompts. Follows #671.
- Logs tab: no issue yet.

Setup flow polish is tracked separately in #757.

jasonmadigan:
@R-Lawton added two issues for what we chatted about

PR #674, opened 2026-07-21
design: add MCP client proxy design doc

R-Lawton:
## Summary

- Defines the proxy architecture for MCP Inspector-like functionality in the console plugin
- Proxy lives in the developer-portal-controller (shared backend for both console plugin and RHDH)
- Covers gateway discovery, auth passthrough, session management, and multi-gateway support

## Key decisions

- Custom proxy needed because MCP servers don't enable CORS and ConsolePlugin spec.proxy is static (can't handle dynamic multi-gateway)
- Proxy discovers gateways from MCPGatewayExtension CRs and routes MCP requests to the correct gateway
- Auth tokens are passed through to the gateway — proxy never stores or validates them
- Session management is stateless in the proxy — frontend holds the Mcp-Session-Id
- Dev preview scope: tools (list, call), prompts (list, get)

## Summary by CodeRabbit

- **Documentation**
  - Added a design document for an integrated MCP inspection and testing experience within OpenShift and RHDH.
  - Documented gateway discovery, endpoint resolution, readiness reporting and MCP request forwarding.
  - Defined authentication and session-header handling, including behaviour for expired or invalid sessions.
  - Recorded security requirements and potential future support for OAuth/OIDC and server-sent notifications.

### DIFF
--- a/docs/designs/2026-07-15-mcp-client-proxy-design.md
+++ b/docs/designs/2026-07-15-mcp-client-proxy-design.md
@@ -1,0 +1,140 @@
+# MCP Client Proxy Design
+
+## Problem
+
+Users of MCP Gateway have no built-in way to inspect, test, or interact with the tools and prompts exposed by the gateway. To test a tool call or browse available prompts, they must use external tools like MCP Inspector, which requires manually entering the gateway URL and runs outside the platform. There's no integrated experience in OpenShift.
+
+## Summary
+
+The MCP client proxy lives in the developer-portal-controller, a shared backend that both the OpenShift console plugin and RHDH already depend on. The proxy is needed because:
+
+- Most MCP servers do not enable CORS — while it can be configured, it is not enabled by default and most servers do not support it. Browsers cannot call them directly without it. MCP Inspector solves the same problem with a Node.js backend.
+- A cluster can have multiple MCP Gateways. The ConsolePlugin CR's `spec.proxy` is static and triggers console pod restarts on changes, so it cannot handle dynamic gateway discovery.
+
+The proxy discovers gateways from MCPGatewayExtension CRs, resolves their endpoints, and makes HTTP requests to them on behalf of the frontend. Users supply their own auth credentials in the UI.
+
+Dev preview scope: tools (list, call), prompts (list, get).
+
+## Goals
+
+- MCP Inspector-like experience for tools (list, call) and prompts (list, get) through the gateway
+- User-supplied auth token, with capability filtering applied by the gateway
+
+## Non-Goals
+
+- SSE notifications (tool list changes, server-initiated events)
+- OAuth/OIDC flow (automatic auth discovery and login)
+
+## Design
+
+### Gateway discovery
+
+The proxy watches MCPGatewayExtension CRs across the cluster using the k8s API. For each CR, it resolves the gateway endpoint:
+
+1. Read `spec.targetRef` on the MCPGatewayExtension — this points to a Gateway resource with a `name`, `namespace`, and `sectionName` (listener)
+2. Read the referenced Gateway resource to get the `gatewayClassName` and the listener's port
+3. If `spec.privateHost` is set on the MCPGatewayExtension, use that as the endpoint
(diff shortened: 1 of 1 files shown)


jasonmadigan:
Will dig into this today

jasonmadigan:
After reading this and some experimentation, I've got a different take on what we should do here.

Taking as given that we want a way to debug/test MCP in-console without reaching for mcp-inspector.*

@R-Lawton 


Some thoughts:

- Re: CORS, most MCP servers won't support it OOTB because they expect to be called by harnesses rather than browsers, and CORS is a browser thing. But that's the wrong hop for this design. Our plugin never calls those servers, it calls the mcp-gateway, and the gateway is ours to configure.
- mcp-gateway already has most of the infra we need for browser-based clients to work. There's a guide for enabling CORS in the [listener/router setup](https://docs.kuadrant.io/1.4.x/mcp-gateway/docs/guides/configure-mcp-gateway-listener-and-router/).
- That guide has bugs that make CORS unusable OOTB, which we should fix regardless of what we decide here:
  - `Access-Control-Expose-Headers` is missing, so there's no way to read `Mcp-Session-Id` after init
  - `Mcp-Session-Id` isn't in `Access-Control-Allow-Headers` either, so preflight fails on every request after `initialize`
  - It pairs `Access-Control-Allow-Origin: *` with `Access-Control-Allow-Credentials: true`, which browsers reject
- To confirm, I fixed those three on a local cluster and ran a full `initialize` / `notifications/initialized` / `tools/list` cycle and stuff worked fine.
- Agree we don't want the embedded inspector able to talk to any host, only gateways we know about. Worth noting that CORS gives us that in both directions: the gateway decides which origins may call it via `allowedOrigins`, and the plugin can only reach gateways it discovered from MCPGatewayExtension. The proxy design goes the other way, `spec.privateHost` is an arbitrary URL the proxy will dial, which is an SSRF surface we'd have to defend (and it would be quite difficult to defend)
- The fix I'd suggest is a `spec.cors` block on MCPGatewayExtension so the controller stamps the headers onto the HTTPRoute it already manages, always including the protocol headers so nobody has to know to add them. Right now the guide's approach requires `httpRouteManagement: Disabled` and hand-maintaining the route per extension, which is probably why nobody's using it
- This scales fine with multiple gateways. You get one extension per namespace and one per Gateway port (not per listener, the EnvoyFilter patches by portNumber), so several extensions means several Gateways plus ReferenceGrants. Each gets its own broker Deployment, Service, EnvoyFilter and controller-managed HTTPRoute in its own namespace. The route is already per-tenant anyway.
- On the auth model: `spec.oauthProtectedResource` already exists on MCPGatewayExtension v1 and carries the authorisation servers, resource, scopes and bearer methods. That's enough for RFC 9728 discovery and Authorisation Code with PKCE from the browser, so we shouldn't need users pasting a bearer token into a text box even for dev preview.

Extending the developer-portal-controller bothered me a bit:

- The proxy would need cluster-wide access to every MCPGatewayExtension and Gateway, plus deeper network access into a plethora of namespaces than we'd probably like, which is a real minefield. OCP network policies for pod-to-pod access would become an issue quickly.
- It also inherits the authorisation problem. The plugin runs on the signed-in user's token today and gets RBAC from the API server for free. A proxy with its own ServiceAccount sees everything, and the design says it validates nothing, so `GET /gateways` would hand the full cluster inventory to any console user.

*I could easily be swayed that we don't want this at all. Is it something we think we want, or has it come from elsewhere? For what it's worth it doesn't look like it originated in this PR: #667 is an accepted epic for MCP management UI with a Figma prototype behind it, and #670-673 hang off that, so someone has designed for it. Worth knowing who's asking before we build the plumbing.

R-Lawton:
hey @jasonmadigan, i think your approach makes a lot of sense here — pushing CORS config into the gateway via `spec.cors` on MCPGatewayExtension is cleaner than the proxy and avoids the security concerns you flagged.

would you mind writing up a design doc for your proposal? i'd like to close this PR in favour of that. we can link them together for context.

jasonmadigan:
I should have a draft design doc for this a little later today. Want to write a small PoC to prove whether some issues with content security policies are real or imaginary...

R-Lawton:
closing in favour of https://github.com/Kuadrant/kuadrant-console-plugin/pull/748

Issue #636, opened 2026-07-13
Merge gateway-api-console-plugin into kuadrant-console-plugin


R-Lawton:
## Context

The [gateway-api-console-plugin](https://github.com/Kuadrant/gateway-api-console-plugin) is a separate console plugin providing Gateway and HTTPRoute create/edit forms and "Attached" resource tabs. Rather than maintaining two plugins, merge its functionality into the kuadrant-console-plugin.

## Subtasks

- [ ] Copy Gateway create/edit form components (`GatewayCreatePage.tsx`, `GatewayModel.tsx`)
- [ ] Copy HTTPRoute create/edit form components (`HTTPRouteCreatePage.tsx`, `HTTPRouteRuleWizard.tsx`, `HTTPRouteModel.tsx`, backend-refs, filters, review)
- [ ] Copy "Attached" tab components (`GatewaySingleOverview.tsx`, `HTTPRouteSingleOverview.tsx`, `AttachedResources.tsx`)
- [ ] Copy HTTPRoute actions (`useHTTPRouteActions.tsx`)
- [ ] Copy utilities (`nameFromPath.ts`, `ParentReferencesSelect.tsx`, `gatewayCreateEditHelpers.ts`)
- [ ] Replace `src/components/gateway/types.ts` stub with full `GatewayResource` type
- [ ] Replace `src/components/httproute/types.ts` stub with full `HTTPRouteResource` type
- [ ] Replace inline `Gateway` interface in `KuadrantOverviewPage.tsx` with `GatewayResource` import
- [ ] Update all references from `name`/`namespace` to `metadata.name`/`metadata.namespace`
- [ ] Migrate PF5 CSS classes to PF6 (`pf-v5-c-*` → `pf-v6-c-*`)
- [ ] Review and update EmptyState usage for PF6 API changes
- [ ] Migrate i18n keys from `plugin__gateway-api-console-plugin` to `plugin__kuadrant-console-plugin`
- [ ] Add console extensions to `console-extensions.json` (Gateway create/edit, HTTPRoute create/edit, Attached tabs, HTTPRoute actions)
- [ ] Drop gateway plugin's `statusLabel.tsx`, switch to existing `getStatusLabel` from `src/utils/statusLabel.tsx`
- [ ] Add i18n translation entries for new strings
- [ ] `npx tsc --noEmit` passes
- [ ] `yarn build` succeeds
- [ ] `yarn lint` passes
[...]

PR #690, opened 2026-07-28
Add gateway api gateway form

R-Lawton:
## Summary
adds Gateway create/edit form and actions from the gateway-api-console-plugin.

part of #636

## What this adds
- Gateway create/edit form (listeners, TLS, allowed routes, addresses)
- Gateway kebab actions (edit labels, edit annotations, edit via form, delete)
- GatewayResource and HTTPRouteResource type consolidation
- KuadrantCreateUpdate redirectPath prop for gateway-api form reuse
- Gateway-api extensions split into separate console-extensions-gateway.json

## Independent from
- #696 (HTTPRoute form) — can be reviewed/merged in parallel

## Testing
- [x] `yarn build` succeeds
- [x] `yarn test` passes
- [ ] Gateway create form renders with listeners wizard
- [ ] Gateway edit form loads existing data
- [ ] Gateway kebab menu shows Edit/Delete actions
- [ ] Form↔YAML sync works both directions

## Summary by CodeRabbit

* **New Features**
  * Added Gateway creation and editing with a guided listener wizard (including TLS, certificates, addresses, and route constraints) plus a YAML editor.
  * Added Gateway actions (edit labels/annotations and delete) and refreshed Gateway/listener/TLS console wording and validation messages.
  * Surfaced additional policy and API product pages, along with related action hooks.

* **Bug Fixes**
  * Improved post-save navigation for policy create/update operations.
  * Updated Gateway traffic status reporting in the overview.

* **Tests**
  * Added end-to-end Gateway CRUD coverage, including form↔YAML synchronisation.


emmaaroche (commented):
Found two bugs while manually testing changes:

**Edit navigates to YAML-only editor instead of form**

Clicking Edit on a Gateway or HTTPRoute (from either the list overview or the resource details page) opens the OpenShift YAML editor, not the new form+YAML toggle view. However editing an HTTPRoute from its resource details page *does* land on the correct form+YAML toggle view. 

**HTTPRoute Create button stays disabled on form view even when all fields are filled**

With route name, parent references, and at least one rule all filled out, the Create button remains disabled. Switching to YAML view enables the Create button.

src/components/gateway/GatewayCreatePage.tsx
```diff
@@ -0,0 +1,1448 @@
...
+      namespaces: {
+        from: 'Same',
+      },
+      kinds: [],
+    },
+  });
+
+  const gatewayWatchResource = React.useMemo(
+    () =>
+      nameEdit
```
emmaaroche: Missing `'~new'` guard. On the create route `nameEdit` is `'~new'`, so it watches for a Gateway literally named `~new` where you'll see a 404 in the network tab on every create page load. 

`HTTPRouteCreatePage` already handles this:

```ts
nameEdit && nameEdit !== '~new'
```

Worth adding the same check here.

Anton-Fil:
Found bugs while manually testing changes Http route form:
1. It's possible to save an incomplete filter if you switch to step 3 (Backend Services) mid-way through configuring a filter on step 2 (Filters). The partially configured filter gets saved without validation.
2. It's not possible to add a second match — the UI only shows Match-1 with no option to add another.
image


3. The wizard is missing a review/summary step before final submission. 
4. When a backend reference is deleted via the YAML view and you switch back to the form view, the reference reappears with no name.

build/suite-router.sh
```diff
@@ -86,7 +86,11 @@ if echo "$CHANGED" | grep -qE "^src/components/(ratelimitpolicy|authpolicy)/"; t
 fi
 
 if echo "$CHANGED" | grep -qE "^src/components/(httproute|issuer)/"; then
-  SPECS="$SPECS rbac.spec.ts"
+  SPECS="$SPECS rbac.spec.ts httproute-crud.spec.ts"
```
emmaaroche, line 89: Would remove this from this PR as already added in https://github.com/Kuadrant/kuadrant-console-plugin/pull/696

src/components/gateway/GatewayCreatePage.tsx
```diff
@@ -0,0 +1,1456 @@
...
+                      from: value as AllowedRoutesUI['namespaces']['from'],
+                    },
+                  },
+                })
+              }
+              aria-label={t('Select Allowed Namespaces')}
+            >
+              <FormSelectOption value="All" label={t('All')} />
+              <FormSelectOption value="Same" label={t('Same')} />
+              <FormSelectOption value="Selector" label={t('Selector')} />
```
emmaaroche, line 1009: Existing bug: `Selector` is available in the `Select Allowed Namespaces` dropdown but there's no input to specify which labels to match. Picking it generates `namespaces: { from: Selector }` with no `namespaces.selector` field. We should add key/value label inputs that show up when `Selector` is picked. We could open an issue for this, or would it be easier to add in this PR, WDYT?

Image

src/components/gateway/GatewayCreatePage.tsx
```diff
@@ -0,0 +1,1456 @@
...
+        populateFormFromGateway(gatewayUpdate);
+        setYamlContent(gatewayUpdate);
+        setIsLoading(false);
+        hasInitializedFromResource.current = true;
+      }
+    } else if (gatewayError) {
+      console.error('Failed to fetch the resource:', gatewayError);
+      setIsLoading(false);
+    }
+  }, [gatewayData, gatewayLoaded, gatewayError, nameEdit]);
```
emmaaroche, line 615: Bug in the hydration effect where after the initial load, any status update from the controller re-triggers the effect. `setIsLoading(true)` fires because `nameEdit` is set, but `setIsLoading(false)` is inside the `!hasInitializedFromResource.current` block which is now `false`, so it never clears. The form gets stuck on the spinner.

Fix is to move `setIsLoading(false)` outside that guard:

```ts
} else {
  if (!hasInitializedFromResource.current) {
    const gatewayUpdate = gatewayData as GatewayResource;
    setCreate(false);
    setOriginalMetadata(gatewayUpdate.metadata);
    populateFormFromGateway(gatewayUpdate);
    setYamlContent(gatewayUpdate);
    hasInitializedFromResource.current = true;
  }
  setIsLoading(false); // moved here
}
```

src/components/gateway/GatewayCreatePage.tsx
```diff
@@ -0,0 +1,1456 @@
...
+            <KuadrantCreateUpdate
+              validation={formValidation()}
+              model={gatewayModel}
+              resource={gatewayObject}
+              policyType="Gateway"
+              navigate={navigate}
+              redirectPath={`/k8s/ns/${selectedNamespace}/${gatewayModel?.apiGroup}~${gatewayModel?.apiVersion}~${gatewayModel?.kind}/${gatewayName}`}
+            />
+          )}
+        </>
```
emmaaroche: UI nit: The Create button stretches full width and there's no Cancel button (not a slay from OG creators),  policy forms wrap `KuadrantCreateUpdate` in an `ActionGroup` with a Cancel button alongside it. 

We could do the same here:

```tsx
  <KuadrantCreateUpdate
    validation={formValidation()}
    model={gatewayModel}
    resource={gatewayObject}
    policyType="Gateway"
    navigate={navigate}
    redirectPath={`/k8s/ns/${selectedNamespace}/${gatewayModel?.apiGroup}~${gatewayModel?.apiVersion}~${gatewayModel?.kind}/${gatewayName}`}
  />
  <Button variant="link" onClick={() => navigate(-1)}>
    {t('Cancel')}
  </Button>
</ActionGroup>
```

Again could do this as part of separate issue if preferred!

emmaaroche (commented):
Looks good, just a few comments below ⬇️

emmaaroche (commented):
Two more things as discussed in office:

1. Clicking Edit on a Gateway from the list overview page opens the OpenShift YAML editor
3. Remove Envoy Gateway class form dropdown option

R-Lawton:
Edit functionality should be working now

emmaaroche (approved):
LGTM

Image

Issue #654, opened 2026-07-14
Bump SDK react-router past CVE-affected 7.13.x


jasonmadigan:
## What

SDK 4.22.0 peers react-router `~7.13.1`, which carries published CVEs:

- **CVE-2026-42211** RCE
- **CVE-2026-53663** CSRF
- **CVE-2026-42342**, **CVE-2026-34077** DoS
- **CVE-2026-40181** open redirect

All fixed in react-router 7.15.1. The console fix is openshift/console#16726, merged 2026-07-16, bumping react-router to `~7.18.1`. No published npm SDK carries it yet: the latest prerelease (`4.23.0-prerelease.4`, 2026-07-14) still peers `~7.13.1`; no `4.22.z` patch exists.

The `__manifest` DoS CVEs reportedly do not apply to the console itself; the concern is scanner findings on the dependency (RHCL downstream scans will flag it).

Track and adopt the first SDK release carrying react-router `>=7.15.1`, whether that is a `4.22.z` patch or `4.23` GA.

Depends on #651.

## Acceptance criteria

- [ ] Local build and scanner dependencies resolve `react-router >=7.15.1` while retaining SDK 4.22 and compatibility with OCP 4.22's host-owned router.
- [ ] Generated federation metadata consumes React Router as a singleton with no bundled fallback.
- [ ] Recursive dependency security scan is clean for `react-router`.

## Update (2026-07-16)

Not currently exploitable via this repo: `yarn.lock` resolves react-router 5.3.4 and 6.22.3 only, and the runtime router comes from the host console image, not our bundle. Exposure begins when #651 moves our SDK peer to react-router 7.x; apply the `>=7.15.1` floor (or a `resolutions` pin) at that point. Remains standalone, not folded into #651.

## Update (2026-08-11)

Two corrections to the original body.

**The 7.15.1 floor is wrong.** react-router 7.13.2 is affected by ten advisories, not the five listed above. A 7.15.1 floor leaves five open:

| advisory | severity | patched in |
|-|-|-|
| GHSA-qwww-vcr4-c8h2 (RSC CSRF bypass) | high | 7.18.2 |
| CVE-2026-55685 DoS via inefficient route matching | high | 7.18.0 |
| CVE-2026-53669 open redirect in `<Link>` and `useNavigate` | medium | 7.18.0 |
| CVE-2026-53667 XSS in RSCErrorHandler | medium | 7.18.0 |
[...]

PR #733, opened 2026-08-11
build(deps): resolve react-router 7.18.2 for dependency scans

jasonmadigan:
## Summary

The plugin bundles no react-router. It is a Module Federation singleton with `allowFallback: false`, so the host console always provides the running copy. On OCP 4.22 that is the console's 7.13.2, before and after this PR. This change is scoped to what our `yarn.lock` reports to dependency scanners; there is no runtime behaviour change.

- Resolve the local react-router devDependency to 7.18.2. It is build-time only (types for the ~37 `src/` files importing `useNavigate`, `useLocation`, `useParams`, `NavigateFunction`).
- Patch SDK 4.22's `react-router` peer range to `>=7.13.1 <8.0.0`. Required, not cosmetic: `ConsoleRemotePlugin` reads that range at build time and fails with `Console provides shared module react-router ~7.13.1 but plugin uses version 7.18.2`. Dependabot #728 is the same bump without the patch and shows exactly this failure. The range also becomes the Module Federation `requiredVersion` for the shared singleton, so it must still accept the 4.22 host's 7.13.x line.
- Verify the generated remote consumes a singleton host router matching `>=7.13.1 <8.0.0`, with no bundled fallback or provider.

7.18.2 is the minimum clean floor, not 7.15.1 as #654 originally stated. Five advisories affecting 7.13.x are only fixed in 7.18.0 or later, including GHSA-qwww-vcr4-c8h2 (7.18.2). Correction filed on #654.

### Why the range is `>=7.13.1 <8.0.0`

A narrow union such as `~7.13.1 || ~7.18.1` would exclude 7.14 to 7.17, so a console shipping one of those would log a federation singleton version warning. The wider range covers every 7.x host and needs no edit on future console bumps.

### Known cosmetic artefact

`yarn install` still prints:

```
YN0060: react-router is listed by your project with version 7.18.2 ...
which doesn't satisfy what @openshift-console/dynamic-plugin-sdk requests (~7.13.1)
```

and `yarn.lock` records the unpatched `~7.13.1` under the patched resolution. Yarn's peer validation reads registry metadata rather than the patched on-disk manifest. The patch is applied; confirm with:

```sh
node -p "require('@openshift-console/dynamic-plugin-sdk/package.json').peerDependencies['react-router']"
# >=7.13.1 <8.0.0
```

## Test evidence

- `yarn install --immutable` — passed
- `yarn lint` and `yarn tsc --noEmit` — passed
- `yarn build` — passed
- `yarn test --runInBand` — 13 suites, 200 tests passed
- `yarn i18n` and `yarn check:spec-map` — passed
- `yarn npm audit --recursive` — no `react-router` advisories
- Built bundle — federation `requiredVersion` is `>=7.13.1 <8.0.0`; no react-router library code and no provide-module in `dist/`

Mitigates #654. That issue stays open to track deleting `.yarn/patches/@openshift-console-dynamic-plugin-sdk-npm-4.22.0-*.patch` once a depended-on SDK release peers react-router 7.18.x directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with OpenShift Console dynamic plugins by updating the supported `react-router` version range.
  * Ensured applications using compatible 7.x router versions resolve correctly.

* **Tests**
  * Added automated checks confirming supported router versions and SDK compatibility.


Anton-Fil:
LGTM

Anton-Fil (dismissed):

R-Lawton (approved):

PR #734, opened 2026-08-11
Gateway UI Bugs Fix

### LINKED ISSUE #708: Gateway create/edit form: multiple UI bugs
---
  Summary

  Several UI bugs found in the Gateway create/edit form (GatewayCreatePage.tsx). Most of these are already solved in the HTTPRoute form and can be fixed by aligning the Gateway implementation with HTTPRoute
  patterns.

  ---
  Bug 1: Listener port becomes NaN when field is cleared

  File: GatewayCreatePage.tsx:703-704

  parseInt("", 10) returns NaN when the user clears the port input. The listener table then shows "NaN" and the resource object is invalid.

  Expected: Fallback to a default value, e.g. parseInt(value, 10) || 80 (as HTTPRoute does).

  ---
  Bug 2: Create/Cancel buttons not visible on small screens
  
  File: GatewayCreatePage.tsx:1519-1533

  On smaller viewports the ActionGroup with Create and Cancel buttons is below the fold and cannot be scrolled to. The form is unusable without these buttons.

  Expected: Buttons should always be reachable — either via a sticky footer or by ensuring the page content is fully scrollable.

Image

  ---
  Bug 3: Wizard modal has broken close button
  
  File: GatewayCreatePage.tsx:1537-1543

  The Modal passes its own title prop which renders a modal header with an X button. The Wizard inside also renders its own header area. The two conflict, resulting in a broken/misplaced close button.

  Expected: Use the HTTPRoute pattern — remove title from Modal and use WizardHeader inside the Wizard component instead:

  <Modal variant="large" isOpen={isModalOpen}>
    <Wizard
      header={
        <WizardHeader
          title={...}
          closeButtonAriaLabel="Close wizard"
          onClose={handleModalClose}
        />
      }
    >

Image

  ---
  Bug 4: Wizard has no step validation — listener can be created with empty required fields

  File: GatewayCreatePage.tsx:1544-1559

  The wizard steps have no validation at all:
  - Next button is always enabled even when required fields (Listener Name, Port) are empty
  - Steps in the sidebar can be jumped to freely regardless of field state
  - Add/Update button on the Review & Create step is always enabled — a listener with Name: "Not specified" can be added

  The HTTPRoute wizard handles this correctly with isDisabled and isNextDisabled on each WizardStep.

  Expected: Disable Next and Add/Update buttons when required fields are not filled. Block jumping to later steps until current step is valid.

Image

pascazzioIsAbyayalan:
## Description

 - Gateway create/edit form had multiple UI bugs: NaN port on clear, buttons unreachable on small screens, broken wizard close button, and no wizard step validation allowing empty listeners to be
  created
- Fixed all four issues in GatewayCreatePage.tsx and added button spacing improvement

  Fixes #708

## Type of change

- [x] `[fix]` Bug fix
- [ ] `[feat]` New feature
- [ ] `[refactor]` Refactor (no functional changes)
- [ ] `[test]` Test updates
- [ ] `[chore]` Dependency / config update

### Port & Buttons

- Added || 0 fallback to parseInt on port input to prevent NaN when field is cleared
- Wrapped ActionGroup in a PageSection with stickyOnBreakpoint={{ default: 'bottom' }} so Create/Cancel are always reachable
- Added marginLeft spacing between Create and Cancel buttons

### Wizard Modal

- Removed title from Modal, replaced with WizardHeader inside the Wizard component to fix the broken close button
- Added isNextDisabled and isDisabled props to each WizardStep to enforce per-step validation
- Configuration step requires non-empty listener name and valid port (1-65535)
- Protocol step requires at least one certificate ref with a name when using HTTPS/TLS Terminate mode
- Later steps are blocked in the sidebar until prior steps are valid

## Test plan

 - [x] yarn build passes
  - [x] yarn lint passes
  - [ ] Clearing the port field shows 0, not NaN
  - [ ] Create/Cancel buttons are visible and reachable on small viewports
  - [ ] Wizard close button (X) works correctly
  - [ ] Wizard Next and Add buttons are disabled when required fields are empty
  - [ ] Wizard sidebar steps cannot be jumped to when prior steps are invalid

## Screenshots

| Buttons | Close button wizard |
|--------|-------|
|gateway-bug2-fix| gateway-bug3-fix|

### Wizard step validation

https://github.com/user-attachments/assets/89f37d19-ea7d-4d71-8033-9159af44d59d

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added stronger validation for gateway names, ports, TLS settings, and route details.
  - Wizard steps now remain unavailable until required information is complete.
  - Added default certificate references for HTTPS and TLS listeners.

- **Bug Fixes**
  - Prevented progression through the gateway creation wizard when entries are invalid.
  - Improved form-change status updates and wizard navigation controls.

### DIFF
--- a/src/components/gateway/GatewayCreatePage.tsx
+++ b/src/components/gateway/GatewayCreatePage.tsx
@@ -17,9 +17,8 @@ import {
   Modal,
   Wizard,
   WizardStep,
-  Tabs,
-  Tab,
-  TabTitleText,
+  WizardHeader,
+  Radio,
   Alert,
   Spinner,
 } from '@patternfly/react-core';
@@ -629,7 +628,10 @@ const GatewayCreatePage: React.FC<GatewayCreatePageProps> = ({ onFormChange }) =
   }, [gatewayObject]);
 
   const onFormChangeRef = React.useRef(onFormChange);
-  onFormChangeRef.current = onFormChange;
+
+  React.useEffect(() => {
+    onFormChangeRef.current = onFormChange;
+  }, [onFormChange]);
 
   React.useEffect(() => {
     if (onFormChangeRef.current) {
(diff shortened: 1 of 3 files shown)


emmaaroche (commented):
All four bugs from #708 addressed, nice work! A few things to wrap up:

1. **Lint CI is failing** - run `yarn lint` locally and commit the auto-fix

2. **E2e gateway CRUD test failing** - the `form↔YAML sync preserves data` test is timing out on a click. Could be related to the sticky PageSection change.

3. **CodeRabbit comments** - could you take a look at the remaining ones? The `onEscapePress` on the Modal is a good catch.

pascazzioIsAbyayalan:
### Solutions Added

**GatewayCreatePage.tsx — CodeRabbit fixes:**

  1. SectionName validation — Listener names must now match the Gateway API SectionName regex and be unique within the Gateway (in both the wizard and final form validation).
  2. Port validation hardened — Added Number.isInteger check to isConfigStepValid.
  3. TLS options row validation — Empty key/value rows in TLS options now block the Protocol step and form submission.
  4. Allowed route kinds validation — Empty kind/group rows now block the Allowed Routes step and form submission.
  5. Prettier formatting — Collapsed the Cancel Button props to one line.
  6. Modal Escape key — Added onEscapePress={handleModalClose} so pressing Escape closes the wizard modal.

**gateway-crud.spec.ts — E2E test fix:**

  7. HTTPS wizard timeout — The addListenerViaWizard helper now adds a certificate ref when protocol is HTTPS/TLS, preventing the "Next" button from being disabled by
  isProtocolStepValid.

.yarnrc.yml
Anton-Fil, line 1: These changes are unrelated to the gateway bug fixes and introduce security-relaxing Yarn settings that aren't mentioned in the PR description:
  
   - approvedGitRepositories: ["**"] — allows Yarn to clone packages from any git repository
   - enableScripts: true — enables lifecycle scripts (postinstall, etc.) which are disabled by default in Yarn 4 for security
   - npmMinimalAgeGate: 0 — disables the age-gate protection against newly published malicious packages
  
   Please revert this file to its original state. If these settings are genuinely needed, they should go in a separate PR with an explanation of why each is required.

package.json
Anton-Fil, line 1: The packageManager bump from 4.12.0 to 4.18.0 and the resulting yarn.lock hash changes are unrelated to the gateway bug fixes. Mixing tooling upgrades with bug fixes makes it harder to bisect regressions —  please revert both package.json and yarn.lock and submit as a separate PR if a Yarn upgrade is needed.

yarn.lock
Anton-Fil, line 1: The packageManager bump from 4.12.0 to 4.18.0 and the resulting yarn.lock hash changes are unrelated to the gateway bug fixes. Mixing tooling upgrades with bug fixes makes it harder to bisect regressions — 
please revert both package.json and yarn.lock and submit as a separate PR if a Yarn upgrade is needed.

src/components/gateway/GatewayCreatePage.tsx
```diff
@@ -671,10 +685,31 @@ const GatewayCreatePage: React.FC = () => {
     return isFormValid;
   };
 
+  const isConfigStepValid =
```
Anton-Fil, line 705: The validation logic in `isConfigStepValid` / `isProtocolStepValid` / `isAllowedRoutesStepValid` (lines 688-706) largely duplicates what's inside `formValidation()` (lines 647-685). If a validation rule changes,
  it needs updating in both places.

  Consider extracting shared predicates, e.g.:
  ```typescript
  const isListenerConfigValid = (l: ListenerUI) =>
    l.name.trim() !== '' &&
    l.name.length <= 253 &&
    SECTION_NAME_RE.test(l.name) &&
    Number.isInteger(l.port) &&
    l.port > 0 &&
    l.port <= 65535;
  ```
  Then reuse in both `formValidation()` and the wizard step guards.

Anton-Fil (changes requested):
looks good overall. Locally, everything works fine. You can slightly enlarge the wizard window to make it easier to read and enter information, but that's your choice.
Image
also some fix:

Anton-Fil (commented):
## Performance & Code Quality

Great work fixing these UI bugs! The validation logic is solid. I have two suggestions:

### 1. Performance - wizardSteps should use React.useMemo

**Location:** `GatewayCreatePage.tsx:709-1245`

The `wizardSteps` array is currently recreated on every render, which happens on every keystroke since it depends on validation state that changes frequently (`isConfigStepValid`, `isProtocolStepValid`, `isAllowedRoutesStepValid`). This is expensive and can cause unnecessary re-renders of the wizard.

**Suggested fix:**
```typescript
const wizardSteps = React.useMemo(() => [
  {
    name: t('Configuration'),
    nextButtonText: t('Next'),
    isNextDisabled: !isConfigStepValid,
    form: (
      // ... form content
    ),
  },
  // ... other steps
], [isConfigStepValid, isProtocolStepValid, isAllowedRoutesStepValid, currentListener, t]);
```

### 2. NIT - Use PatternFly spacing utility instead of inline style

**Location:** `GatewayCreatePage.tsx:1576`

The Cancel button uses inline style `style={{ marginLeft: '24px' }}`. We should prefer PatternFly utility classes for consistency:

```typescript
<Button variant="link" onClick={() => navigate(-1)} className="pf-v6-u-ml-md">
  {t('Cancel')}
</Button>
```

Or if you need the exact spacing, use the PatternFly variable:
```typescript
style={{ marginLeft: 'var(--pf-v6-global--spacer--md)' }}
```

R-Lawton:
nice cleanup on the validation refactor. looks like you already extracted the helpers anton suggested and dropped the yarn/`.yarnrc.yml` changes, so those are sorted.

one thing — the description says bug #3 (broken wizard close button) was fixed by removing the modal title and adding a `WizardHeader`, but i don't see that in the diff. the `Modal` still uses `title` and there's no `WizardHeader`. is that fix actually in here, or did it get handled somewhere else? wanna make sure the close button bug is genuinely covered before merge.

minor: line 691 `certificateRefs.every((ref) => ref.name !== '')` in `formValidation` looks redundant now that `isListenerProtocolValid` covers it (and it's stricter with `.trim()`). could drop it, wdyt?

pascazzioIsAbyayalan:
> nice cleanup on the validation refactor. looks like you already extracted the helpers anton suggested and dropped the yarn/`.yarnrc.yml` changes, so those are sorted.
> 
> one thing — the description says bug #3 (broken wizard close button) was fixed by removing the modal title and adding a `WizardHeader`, but i don't see that in the diff. the `Modal` still uses `title` and there's no `WizardHeader`. is that fix actually in here, or did it get handled somewhere else? wanna make sure the close button bug is genuinely covered before merge.
> 
> minor: line 691 `certificateRefs.every((ref) => ref.name !== '')` in `formValidation` looks redundant now that `isListenerProtocolValid` covers it (and it's stricter with `.trim()`). could drop it, wdyt?

I agree the `certificateRefs.every` check is redundant with the use of `.trim()`, will remove. I noticed I also missed adding the `WizardHeader`. I will get on that right away :relaxed:

pascazzioIsAbyayalan:
@Anton-Fil 

1. Fixed React Doctor error (ref mutation)
2. Performance optimization (useMemo)
3. Code quality (PatternFly utilities)

@R-Lawton 

1. Bug # 3 fixed (wizard close button)
2. Removed redundant validation

CodeRabbit suggestions related to this Issue have been addressed. Findings that affect other sections will have a follow up soon :dizzy:

pascazzioIsAbyayalan:
Bug #3 fix + wizard hight 
Screenshot 2026-08-19 at 00 25 04

emmaaroche (commented):
**src/components/gateway/GatewayCreatePage.tsx:671-675**

TLS options validation runs for all listeners, even HTTP. If user adds incomplete TLS option then switches protocol to HTTP, wizard stays blocked - inputs hidden but predicate still false. Options won't persist anyway (cleanup in `handleListenerSave` line 190).

Move `l.tlsOptions.every(...)` check inside TLS condition:

```typescript
const isListenerProtocolValid = (l: ListenerUI) =>
  (l.protocol !== 'HTTPS' && l.protocol !== 'TLS') ||
  l.tlsMode !== 'Terminate' ||
  (l.certificateRefs.length > 0 &&
    l.certificateRefs.every((ref) => ref.name.trim() !== '') &&
    l.tlsOptions.every((opt) => opt.key.trim() !== '' && opt.value.trim() !== ''));
```

This way HTTP/UDP/TCP listeners skip TLS checks entirely, and Passthrough mode doesn't validate options that get stripped anyway.

emmaaroche:
Is this rebased yet? 😄

Anton-Fil:
look like neeed rebase

pascazzioIsAbyayalan:
Give me a second :tired_face:

src/components/gateway/GatewayCreatePage.tsx
```diff
@@ -1233,335 +1297,354 @@ const GatewayCreatePage: React.FC<GatewayCreatePageProps> = ({ onFormChange }) =
...
+                <Button variant="link" onClick={() => navigate(-1)} className="pf-v6-u-ml-md">
+                  {t('Cancel')}
+                </Button>
+              </ActionGroup>
+            </PageSection>
+          )}
+        </>
+      )}
+
+      <Modal variant="large" isOpen={isModalOpen}>
```
emmaaroche: The `Modal` doesn't have `onClose` or `onEscapePress` wired up, so pressing Escape won't close the listener wizard - only the `WizardHeader`'s X button works. CodeRabbit's earlier comment on this got marked "addressed," but it looks like that fix didn't survive a rebase/force-push since the current branch head doesn't have it. Could we add `onEscapePress={handleModalClose}` here?

src/components/gateway/GatewayCreatePage.tsx
```diff
@@ -656,566 +658,628 @@ const GatewayCreatePage: React.FC<GatewayCreatePageProps> = ({ onFormChange }) =
...
-            >
+  const isConfigStepValid =
+    isListenerConfigValid(currentListener) &&
+    !listeners.some((l, i) => i !== editingListenerIndex && l.name === currentListener.name);
+
+  const isProtocolStepValid = isListenerProtocolValid(currentListener);
+
+  const isAllowedRoutesStepValid = isListenerAllowedRoutesValid(currentListener);
+
+  const wizardSteps = React.useMemo(
```
emmaaroche: `wizardSteps` is wrapped in `React.useMemo`, but the dependency array includes handlers like `handleAddCertificateRef`, `handleCertificateRefChange`, etc. (lines 1268-1278) that aren't wrapped in `useCallback` - they're plain function declarations recreated on every render. Since the dependency array identity changes every render anyway, the memo never actually skips recomputation. Either wrap those handlers in `useCallback`, or drop the `useMemo` since it's not buying anything right now.

emmaaroche (approved):
LGTM 💃

Anton-Fil (approved):
LGTM

Issue #656, opened 2026-07-16
Bump consolePlugin latestSupportedOpenshiftVersion to 4.22


jasonmadigan:
## What

Plugin metadata still declares `consolePlugin.latestSupportedOpenshiftVersion: "4.19"` in `package.json`. The plugin is tested on OCP 4.22 and expected to be 4.23/5.0-compatible via the console's `react-router-dom-v5-compat` alias (verification tracked in #653), so the declared ceiling is stale. This metadata ships in the RHCL 1.5 build unless bumped. Relates to #650 and #653.

## Acceptance criteria

- [ ] `latestSupportedOpenshiftVersion` bumped to `"4.22"` in the `package.json` `consolePlugin` stanza
- [ ] Built plugin manifest carries the new value

jasonmadigan:
Draft PR #733 created: https://github.com/Kuadrant/kuadrant-console-plugin/pull/733

jasonmadigan:
Completed by merged PR #721: `consolePlugin.latestSupportedOpenshiftVersion` is 4.22, the manifest requires `@console/pluginAPI >=4.22.0-0`, and manual and E2E verification passed on OCP 4.22.

Issue #653, opened 2026-07-14
Test plugin against OCP 4.23/5.0 PatternFly 6.6 themes


jasonmadigan:
## What

OCP 4.23/5.0 consoles upgrade PatternFly CSS from 6.4.x to 6.6.x, introducing two new themes: glass and high-contrast (per the SDK 4.23 release notes). The console team's plugin test window (#tmp-ocp-pf6-upgrade clusters) ran to 21 May 2026 and we did not participate.

Verify every view in glass and high-contrast alongside light and dark, and fix regressions. Theme-specific CSS selectors in the codebase (topology dark-theme rules in `src/components/kuadrant.css` lines 18-66, `RejectionModal.css`, `ApprovalModal.css`) are the likely problem areas.

## Acceptance criteria

- [ ] All pages checked in glass, high-contrast, light and dark modes on an OCP 4.23/5.0 console build
- [ ] Regressions fixed with kuadrant-scoped CSS only

## Update (2026-07-16)

Dependency on #651 dropped: the build shipping to 4.23/5.0 consoles in RHCL 1.5 is the current v5-compat build, so theme testing targets it directly and should run before the October cut. Testable now: `start-console.sh` defaults to `quay.io/openshift/origin-console:latest`, which tracks console master (4.23/5.0 dev); check the theme work has merged into that image first. Align terminology with CONNLINK-1206 ([RHCL] PatternFly 6 and Compass readiness); Compass is a separate PatternFly effort, distinct from these two themes.

jasonmadigan:
Draft PR #735: https://github.com/Kuadrant/kuadrant-console-plugin/pull/735

Issue #651, opened 2026-07-14
Migrate to @openshift-console/dynamic-plugin-sdk 4.22.0


jasonmadigan:
## What

Bump `dynamic-plugin-sdk` `^1.6.0` and `dynamic-plugin-sdk-webpack` `1.2.0` to matched 4.22.0 releases. The two must move together: a version mismatch fails at runtime with `__load_plugin_entry__ is not defined` (`loadPluginEntry` was removed).

This is atomic with React 18 (#637) and react-router 7 (#333): the console's shared modules move together in 4.22, so these cannot land separately on main. Router usage across the 29 files importing `react-router-dom-v5-compat` is only `useNavigate`/`useLocation`/`useParams`/`Link`/`NavigateFunction`, all identical in react-router 7, so the migration is an import-path swap plus removing `react-router`, `react-router-dom` and `react-router-dom-v5-compat` from `package.json`. React 18 scope and codebase scan are in #637.

Type fallout to expect: `K8sResourceCommon`/`ObjectMetadata`/`OwnerReference` change from type aliases to interfaces; `useDeleteModal` `redirectTo` takes the react-router `To` type. All console extensions we use survive 4.22 and `ConsoleRemotePlugin()` takes no options, so no `transformImports`/`moduleFilter` change needed.

SDK 4.22.0 peers react-router `~7.13.1`, which carries published CVEs (CVE-2026-42211 RCE, CVE-2026-53663 CSRF, CVE-2026-42342 and CVE-2026-34077 DoS, CVE-2026-40181 open redirect, all fixed in 7.15.1). The fix (openshift/console#16726, bumping to `~7.18.1`) merged 2026-07-16; no published npm SDK carries it yet (latest `4.23.0-prerelease.4` still peers `~7.13.1`). Tracked in #654, not a blocker here.

Risks: `react-helmet` (`^6.1.0`, devDependencies, console shared module) and `swagger-ui-react` (`5.10.5`, regular dependency) on React 18.

Blocked by #647, #648, #649, #650. Covers #637 and #333.

Reference migrations: kubevirt-ui/kubevirt-plugin#3825, openshift/networking-console-plugin#371, openshift/nmstate-console-plugin#188.

The `/console:upgrade-console-sdk` Claude Code skill can assist with the SDK bump and associated type fixes.

## Checked, not applicable

[...]

Issue #650, opened 2026-07-14
Cut release-0.x branch for OCP <=4.21 support


jasonmadigan:
## What

A plugin built on SDK 4.22 (React 18, react-router 7 shared modules) will not run on older consoles. `react` and `react-router` are federation singletons with `allowFallback: false`, so a plugin compiled against React 18/router 7 gets the v17/v5 singletons on <=4.21 consoles and breaks at runtime.

Before the SDK bump merges, cut `release-0.x` from the last pre-bump commit. Main serves OCP 4.22+; `release-0.x` takes backports. This adapts the per-OCP-version branching model used by `kubevirt-plugin` and `networking-console-plugin` (they branch `release-4.y`; ours is `release-0.x` because we version independently of OCP). The branch caps `consolePlugin.latestSupportedOpenshiftVersion` at `4.21` (currently `4.19` with `pluginAPI: "*"` on main). Helm chart and CI publish distinct image tags per stream.

## Acceptance criteria

- [ ] `release-0.x` cut before the SDK 4.22 bump merges
- [ ] Branch declares `latestSupportedOpenshiftVersion: "4.21"`
- [ ] CI/image builds publish distinct tags per stream
- [ ] Backport policy noted in docs

## Update (2026-07-16)

The RHCL 1.5 OCP floor is unpublished: supported-configs article 7092611 (updated 2026-06-25) lists RHCL 1.4 = OCP 4.19-4.22 and has no 1.5 row. Cut timing is gated on that floor; confirm with product management. Two viable strategies: cut `release-0.x` when #651 lands, or hold #651 off main until the RHCL 1.5 downstream branch is cut. They only diverge if the floor stays at 4.20/4.21. The branch also serves community users on <=4.21, independent of RHCL. Which branch RHCL builds consume is owned midstream (this repo carries no downstream build config, only the `downstream.js` rebrand script); confirm with the midstream owner.

PR #736, opened 2026-08-11
docs: versioning and OpenShift compatibility

jasonmadigan:
## Summary

- Replace the stale README matrix with the images to use before and after OCP 4.22.
- Document why the 4.22 build needs a separate stream, and keep the branch and operator maintenance notes in `RELEASE.md`.

## Test evidence

`git diff --check` passes. Documentation only.

Relates to #650, #651, #653, #654, #656 and #737.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ Based on https://github.com/openshift/console-plugin-template
 ## Screenshots
 
 ![Overview](docs/images/overview.gif)
+
 ## Running
 
 - Target a running OCP with `oc login`
@@ -101,6 +102,9 @@ Two easy ways to deploy.
 
 ### Via `kubectl` or `oc`
 
+`install.yaml` uses the 4.22+ `latest` image. For OCP 4.21 and earlier, replace
+it with `quay.io/kuadrant/console-plugin:v0.6.0` before applying the manifest.
+
 `oc apply -f install.yaml`
 
 or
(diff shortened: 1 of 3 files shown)


R-Lawton:
this is a lot of info and repeats itself a lot. a lot of this isn't something users of the doc would necessarily need to know — i'd remove these sections to make it more readable wdyt?:

- "Where v0.6.0 sits"
- the paragraph about why Kuadrant/kuadrant-operator#637, Kuadrant/kuadrant-operator#651 and Kuadrant/kuadrant-operator#333 had to merge as one change
- "the 4.22 stack on main has no release tag yet" paragraph
- "Adopting a new console generation"
- "Choosing an SDK version"
- "Known gaps"

i think some of this is useful but belongs in a different doc — backports, security bumps, and choosing an SDK version might fit better in the release doc?

jasonmadigan:
agreed, this is pretty wordy. I'll trim and rehome most of this in RELEASE.md - I'll try to keep some of the mechanical stuff that might help guide an agent with upgrades in future

the real meat here is probably:

>= 4.22 folks: use img-x
<= 4.21 folks: use img-y

R-Lawton:
noticed a few more things:

- `.yarn/patches` directory doesn't exist in the repo — the text references it but it should note it needs to be created, or describe the `yarn patch` workflow
- the quay cleanup script (`scripts/delete-quay-tags.sh`) would delete `release-0.x-latest` tags since they don't match the keep pattern — worth calling out rather than just saying "check that quay cleanup covers both branches"
- there's no guidance on cutting a release from `release-0.x` (e.g. `v0.6.1`) and `unit-tests.yaml` on `main` only triggers for PRs targeting `main`, so PRs targeting `release-0.x` don't run unit tests — mind creating an issue to track these?

R-Lawton (approved):
approve follow ups can be done later

PR #755, opened 2026-08-18
fix(overview): show successful requests and error rate when no errors exist

### LINKED ISSUE #724: Traffic Analysis: Successful Requests and Error Rate show blank when no errors exist
## Summary

The Traffic Analysis view shows `-` for **Successful Requests** and **Error Rate** when all requests are successful (i.e., the error query returns an empty vector from Prometheus). The values should display the actual request count and `0%` respectively.

## Steps to reproduce

1. Deploy RHCL with a working gateway, AuthPolicy, and RateLimitPolicy (all Accepted/Enforced)
2. Configure observability with User Workload Monitoring and a ServiceMonitor for the gateway
3. Generate traffic — all requests return 2xx responses (no errors)
4. Open the RHCL console plugin → Traffic Analysis

**Expected:** Successful Requests = total request count, Error Rate = 0%
**Actual:** Successful Requests = `-`, Error Rate = `-`

## Root cause

In `KuadrantOverviewPage.tsx`, the error request query returns an empty vector when there are no errors. The plugin never populates `totalRequestsByGateway[key].errors`, so it is `undefined`.

```ts
// Line 495
const success = totalRequestsByGateway[key]?.total - totalRequestsByGateway[key]?.errors;
// total - undefined = NaN
return Number.isFinite(success) ? Math.round(success) : 0;
// → returns 0
```

The JSX then renders:
```tsx
{getSuccessfulRequests(obj) || '-'}
// 0 || '-' = '-'  ← 0 is falsy in JS
```

The same issue affects Error Rate (`NaN / total = NaN → '-'`).

## Fix

Treat `undefined` errors as `0`:

```ts
const errors = totalRequestsByGateway[key]?.errors ?? 0;
const success = totalRequestsByGateway[key]?.total - errors;
```

Or adjust the render to use `!== undefined` rather than a falsy check for the success count.

R-Lawton:
## What

Fixes #724. On the Kuadrant Overview -> Traffic Analysis table, **Successful Requests** and **Error Rate** showed `-` when a gateway had traffic but no errors. They now show the real count and `0.0%`.

## Root cause

When all requests succeed, Prometheus returns an empty error vector, leaving `errors` undefined. `total - undefined` -> `NaN`, and the render used a falsy check (`0 || '-'`), so a valid `0` rendered as `-`.

## Fix

- Treat a missing/empty error vector as `0` errors.
- Use an explicit `!== null` check when rendering (0 is a valid value).
- Extract the derivation logic into a pure `src/utils/gatewayTraffic.ts` util + unit tests for the empty-error-vector case.

## Testing

- `gatewayTraffic.test.ts`: 14 tests (incl. #724 regressions) pass
- Existing `KuadrantOverviewPage.test.tsx`: 2 pass
- `tsc --noEmit` + `eslint`: clean
- Verified live in the console (before -> after, same input):

| | Total | Successful | Error Rate |
|---|---|---|---|
| Before | 1000 | `-` | `-%` |
| After | 1000 | 1000 | 0.0% |

verify-724-after

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gateway traffic metrics for total requests, successful requests and error rates.
  * Correctly distinguish missing metrics from genuine zero values.
  * Display error rates with percentage symbols and consistent rounding.
  * Treat absent error data as zero, including successful-request calculations.

### DIFF
--- a/src/components/KuadrantOverviewPage.tsx
+++ b/src/components/KuadrantOverviewPage.tsx
@@ -69,6 +69,13 @@ import {
 import { fetchConfig, KuadrantConfig } from '../utils/configLoader';
 import { GatewayResource } from './gateway/types';
 import { getStatusSortRank } from '../utils/statusLabel';
+import {
+  GatewayTrafficMetrics,
+  getTotalRequests as getTotalRequestsFor,
+  getSuccessfulRequests as getSuccessfulRequestsFor,
+  getErrorRateValue as getErrorRateValueFor,
+  formatErrorRate as formatErrorRateFor,
+} from '../utils/gatewayTraffic';
 
 export type MenuToggleElement = HTMLDivElement | HTMLButtonElement;
 
@@ -94,13 +101,7 @@ export const resources: Resource[] = [
 ];
 
 interface TotalRequestsByGatewayResource {
-  [gatewayName: string]: {
-    total?: number;
-    errors?: number;
-    codes?: {
-      [responseCode: string]: number;
-    };
-  };
+  [gatewayName: string]: GatewayTrafficMetrics;
 }
 
 type ComputedSortValue = number | string | null | undefined;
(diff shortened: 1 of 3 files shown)


emmaaroche:
👀

src/components/KuadrantOverviewPage.tsx
```diff
@@ -595,8 +584,13 @@ const KuadrantOverviewPage: React.FC = () => {
   };
   const gatewayTrafficRenders = {
     totalRequests: (_column, obj) => getTotalRequests(obj) || '-',
```
emmaaroche: Same bug as the `successfulRequests` fix: `getTotalRequests(obj) || '-'` shows `-` for a genuinely-zero total, not just missing data. Maybe make `getTotalRequests` return `null` when there's no metrics entry, and check `!== null` here too, like the other two renderers.

R-Lawton: good catch

emmaaroche (approved):
LGTM!

PR #746, opened 2026-08-17
LFX Mentorship: Feat: Implement RateLimitPolicy view

justin212407:
<html><head></head><body><h2>Description</h2>
<p>Fixes #378</p>
<p>Adds a form-based creation/editing view to <code>KuadrantRateLimitPolicyCreatePage.tsx</code>,
which previously only rendered a static YAML editor with a hardcoded example resource.
Users can now create and edit RateLimitPolicy resources through a guided form interface
without needing to know the CRD spec structure, while retaining the option to switch
to a YAML view at any point.</p>
<p>This is the fourth deliverable in the
<a href="https://claude.ai/chat/66cb4e94-319c-4820-a2bd-0b1206e9a43d#378">Add form views for policy creation pages</a> epic. RateLimitPolicy is the most
complex policy form so far — it reuses the existing <code>AddLimitModal</code> and <code>LimitSelect</code>
components from <code>src/components/ratelimitpolicy/</code> and supports a deeply nested
<code>spec.limits</code> structure with named limit entries, each containing rates with counters
and conditions.</p>
<h2>Type of change</h2>
<ul>
<li>[x] <code>[feat]</code> New feature</li>
</ul>
<h2>Changes made</h2>
<ul>
<li>Replaced static YAML-only editor in <code>KuadrantRateLimitPolicyCreatePage.tsx</code> with
a full form view following the five-pillar architecture from
<code>KuadrantDNSPolicyCreatePage.tsx</code></li>
<li>Added Form/YAML toggle (PatternFly Radio buttons) — both views read from the same
underlying state</li>
<li>Implemented <code>createRateLimitPolicy()</code> factory function that builds the complete K8s
resource object from form state, used as input to both <code>ResourceYAMLEditor</code> and
<code>KuadrantCreateUpdate</code></li>
<li>Implemented bidirectional YAML sync — <code>useEffect</code> updates YAML when form fields
change; <code>handleYAMLChange()</code> parses edited YAML back into form state via <code>js-yaml</code></li>
<li>Added edit mode hydration — URL path detection triggers <code>useK8sWatchResource</code> to
load existing resource; <code>useEffect</code> populates all form fields including limits;
name field locked via <code>formDisabled</code></li>
<li>Reused <code>LimitSelect</code> and <code>AddLimitModal</code> components from
<code>src/components/ratelimitpolicy/</code> for the limits UI — these handle the nested
<code>spec.limits</code> structure with named entries, rates, counters, and conditions</li>
[...]

PR #774, opened 2026-08-25
feat: add Gateway/HTTPRoute/GRPCRoute target selection to remaining policy forms

### LINKED ISSUE #772: Add HTTPRoute/GRPCRoute target selection to remaining policy forms
## Summary
Extend the "Target Type" selector (Gateway/HTTPRoute/GRPCRoute) being added to the RateLimitPolicy form in #746 to the other policy forms that support the same target kinds.

## Motivation
Per `RESOURCE_POLICY_MAP` in `src/utils/resources.ts`, TokenRateLimitPolicy, OIDCPolicy, PlanPolicy, and AuthPolicy can all target Gateway, HTTPRoute, and GRPCRoute — but their forms currently only expose a `GatewaySelect`. Users targeting an HTTPRoute or GRPCRoute with one of these policies have to drop to YAML, even though the form otherwise supports their use case.

## Proposed Solution
Reuse the pattern from #746: a `targetRef`-as-single-source-of-truth state, a "Target Type" radio group, and conditional rendering of `GatewaySelect` / `HTTPRouteSelect` (`kind="HTTPRoute"`) / `HTTPRouteSelect` (`kind="GRPCRoute"`).

- `KuadrantTokenRateLimitPolicyCreatePage.tsx`
- `KuadrantOIDCPolicyCreatePage.tsx`
- `KuadrantPlanPolicyCreatePage.tsx`
- AuthPolicy's form, once it exists (currently YAML-only)

DNSPolicy and TLSPolicy are out of scope — those CRDs only support targeting Gateway.

## Dependencies
Depends on #746 merging first (introduces the `targetRef` pattern and generalizes `HTTPRouteSelect` to accept a `kind` prop). Part of the "Add form views for policy creation pages" epic (#378).

emmaaroche:
## Description
Closes #772. Extends the Target Type selector pattern introduced in #746 to the remaining policy forms that support targeting a Gateway, HTTPRoute, or GRPCRoute — previously these forms only exposed a `GatewaySelect` (or, for PlanPolicy, a hardcoded HTTPRoute selector), forcing users onto YAML to target a Route.

Also carries over two fixes needed for #746's RateLimitPolicy YAML page:
- `KuadrantRateLimitPolicyCreatePage.tsx` was missing the `gateway-api-plugin.css` import entirely, so the YAML tab's Monaco editor had no height rules and collapsed to 0px.
- `.kuadrant-ratelimitpolicy-yaml-editor` was missing from the shared height selector blocks in `gateway-api-plugin.css`.
- Also dropped `targetRef.namespace` from the RateLimitPolicy save payload — the CRD's `targetRef` doesn't support cross-namespace targeting, so sending it tripped the admission webhook (`unknown field "spec.targetRef.namespace"`).

## Changes
- `KuadrantTokenRateLimitPolicyCreatePage.tsx`, `KuadrantOIDCPolicyCreatePage.tsx`, `KuadrantPlanPolicyCreatePage.tsx`: replaced the Gateway-only (or HTTPRoute-only) target state with `targetRef` as the single source of truth, added the Target Type radio group (Gateway/HTTPRoute/GRPCRoute), and conditionally render `GatewaySelect` / `HTTPRouteSelect`
- `KuadrantRateLimitPolicyCreatePage.tsx`: added missing `gateway-api-plugin.css` import, removed `targetRef.namespace` from the save payload
- `gateway-api-plugin.css`: added `.kuadrant-ratelimitpolicy-yaml-editor` to the shared yaml-editor height selector blocks
- Removed the now-unused `"Target HTTPRoute"` locale key (superseded by the generic Target Type labels)

## Testing
- `yarn lint` and `yarn i18n` pass
- `npx tsc --noEmit` clean on all touched files
- Manually verified Target Type switching and save for Gateway/HTTPRoute/GRPCRoute on all three forms

## Summary by CodeRabbit

- **New Features**
  - OIDC, Plan, and Token Rate Limit policy forms now support Gateway, HTTPRoute, and conditionally GRPCRoute targets.
  - Added target-type selection with context-sensitive resource pickers.
  - Target selections are retained when editing policies or working with YAML.
  - Gateway and HTTPRoute selection is scoped to the selected namespace across policy forms.
- **Bug Fixes**
  - Unsupported YAML target references are rejected instead of being silently accepted.
  - Existing policy target controls are now protected from unintended changes.
  - Corrected Rate Limit policy target references by omitting unsupported namespace data.
- **Style**
  - Improved layout and sizing for Rate Limit policy YAML editors.


src/components/KuadrantOIDCPolicyCreatePage.tsx
```diff
@@ -228,7 +264,51 @@ const KuadrantOIDCPolicyCreatePage: React.FC = () => {
...
+                  <Radio
+                    name="target-type-radio"
+                    label={t('HTTPRoute')}
+                    id="target-type-radio-httproute"
+                    isChecked={targetRef.kind === 'HTTPRoute'}
+                    onChange={() => handleTargetTypeChange('HTTPRoute')}
+                  />
+                  <Radio
+                    name="target-type-radio"
+                    label={t('GRPCRoute')}
```
jasonmadigan: OIDCPolicy only permits Gateway and HTTPRoute targets, and TokenRateLimitPolicy has the same restriction; both CRDs reject GRPCRoute, and the console’s `RESOURCE_POLICY_MAP` excludes these policies from it. These new options pass the form validation and then fail admission. Could we derive the allowed kinds from the shared policy mapping, derive the group from `RESOURCES.Gateway.gvk.group`, and remove GRPCRoute from the OIDC and TokenRateLimit forms?

src/components/KuadrantRateLimitPolicyCreatePage.tsx
```diff
@@ -78,7 +79,6 @@ const KuadrantRateLimitPolicyCreatePage: React.FC = () => {
           group: targetRef.group,
           kind: targetRef.kind,
           name: targetRef.name,
```
jasonmadigan, line 81: Dropping `namespace` from the payload matches the local target-reference schema, but `GatewaySelect` still lists Gateways across namespaces. Selecting `other-ns/foo` here serializes a reference to `policy-ns/foo`, which can attach to a different same-named Gateway or produce `TargetNotFound`. Could we scope the selector to the policy namespace, including a concrete namespace requirement in `#ALL_NS#` mode, before discarding the selected namespace? The new Gateway adapters in the other forms need the same constraint.

src/components/KuadrantPlanPolicyCreatePage.tsx
```diff
@@ -54,9 +65,10 @@ const KuadrantPlanPolicyCreatePage: React.FC = () => {
   const [createView, setCreateView] = React.useState<'form' | 'yaml'>('form');
   const [policyName, setPolicyName] = React.useState('');
   const [selectedNamespace] = useActiveNamespace();
-  const [selectedRoute, setSelectedRoute] = React.useState<{ name: string; namespace: string }>({
+  const [targetRef, setTargetRef] = React.useState<TargetRef>({
+    group: GATEWAY_API_GROUP,
+    kind: 'Gateway',
```
jasonmadigan: This changes PlanPolicy’s initial target from HTTPRoute to Gateway, and the smoke job is now failing both PlanPolicy tests because `#httproute-select` is no longer rendered initially. Could we preserve HTTPRoute as the default? If Gateway is an intentional product change, the tests need to select HTTPRoute explicitly and the changed default should be called out.

src/components/KuadrantPlanPolicyCreatePage.tsx
```diff
@@ -260,14 +305,53 @@ const KuadrantPlanPolicyCreatePage: React.FC = () => {
...
+                  isInline
+                  fieldId="target-type-radio-group"
+                  label={t('Target Type')}
+                >
+                  <Radio
+                    name="target-type-radio"
+                    label={t('Gateway')}
+                    id="target-type-radio-gateway"
+                    isChecked={targetRef.kind === 'Gateway'}
+                    onChange={() => handleTargetTypeChange('Gateway')}
```
jasonmadigan, line 341: In edit mode, `formDisabled` only reaches the route selectors; the new radios and `GatewaySelect` remain enabled. Switching type clears the target name and can leave a disabled empty route picker, while choosing Gateway permits retargeting. Could we make target editability consistent—probably by disabling the radios and every selector for an existing PlanPolicy?

src/components/KuadrantOIDCPolicyCreatePage.tsx
```diff
@@ -155,12 +166,11 @@ const KuadrantOIDCPolicyCreatePage: React.FC = () => {
...
       setPolicyName(parsedYaml.metadata?.name || '');
-      setSelectedGateway({
-        metadata: {
-          name: parsedYaml.spec?.targetRef?.name || '',
-          namespace: parsedYaml.metadata?.namespace ?? '',
-        },
-      } as GatewayResource);
+      setTargetRef({
+        group: parsedYaml.spec?.targetRef?.group || GATEWAY_API_GROUP,
+        kind: (parsedYaml.spec?.targetRef?.kind as TargetRef['kind']) || 'Gateway',
```
jasonmadigan: This assertion does not validate the YAML value at runtime. An unsupported group or kind can be copied into Form state, satisfy the name-only target validation, and then fail admission; the PlanPolicy and TokenRateLimitPolicy pages use the same pattern. Could we validate YAML-to-Form imports against each policy’s supported targets and surface an invalid Form state, while leaving raw YAML submission unchanged?

src/components/KuadrantOIDCPolicyCreatePage.tsx
```diff
@@ -228,7 +264,51 @@ const KuadrantOIDCPolicyCreatePage: React.FC = () => {
...
+                    id="target-type-radio-gateway"
+                    isChecked={targetRef.kind === 'Gateway'}
+                    onChange={() => handleTargetTypeChange('Gateway')}
+                  />
+                  <Radio
+                    name="target-type-radio"
+                    label={t('HTTPRoute')}
+                    id="target-type-radio-httproute"
+                    isChecked={targetRef.kind === 'HTTPRoute'}
+                    onChange={() => handleTargetTypeChange('HTTPRoute')}
```
jasonmadigan, line 308: Could we add passing E2E coverage that selects a non-default target in each independently implemented form and asserts the persisted `spec.targetRef.kind` and `name`? A focused nightly case for switching/reset behaviour and YAML-to-Form hydration would cover the new state logic; the existing tests do not exercise these radios.

jasonmadigan (changes requested):

src/components/KuadrantOIDCPolicyCreatePage.tsx
```diff
@@ -173,23 +196,49 @@ const KuadrantOIDCPolicyCreatePage: React.FC = () => {
...
+          namespace: targetRef.kind === 'Gateway' ? selectedNamespace : '',
+        },
+      } as GatewayResource),
+    [targetRef, selectedNamespace],
+  );
+
+  const selectedRoute = React.useMemo(
+    () => ({
+      name: targetRef.kind === 'Gateway' ? '' : targetRef.name,
+      namespace: targetRef.kind === 'Gateway' ? '' : selectedNamespace,
```
jasonmadigan, line 240: after selecting `ns-a/foo`, changing the active namespace leaves `targetRef.name` as `foo`, while this memo relabels it as `ns-b/foo`. the form remains valid and can target a different route or produce `TargetNotFound`. could we retain the selected route’s original namespace and clear or reject it whenever `selectedNamespace` changes, including `#ALL_NS#`? the Plan and Token forms have the same pattern.

src/utils/resources.ts
```diff
@@ -425,6 +425,25 @@ export const getPoliciesForResource = (resourceKind: ResourceKind): ResourceKind
...
+export const getTargetKindsForPolicy = (policyKind: ResourceKind): ResourceKind[] => {
+  return (Object.keys(RESOURCE_POLICY_MAP) as ResourceKind[]).filter((targetKind) =>
+    RESOURCE_POLICY_MAP[targetKind]?.includes(policyKind),
+  );
+};
+
+// check whether a targetRef (as parsed from user-supplied YAML) is one this
+// policy kind's CRD actually accepts, so unsupported references never get
+// past form state into a save that the admission webhook will reject
+export const isSupportedTargetRef = (
```
jasonmadigan, line 441: this validates only the group and kind. a YAML value containing `targetRef.namespace: other-ns` passes, then the Form builders discard that namespace and silently reinterpret the name as a local target. could we validate the complete target reference during YAML-to-Form hydration and reject fields the Form cannot round-trip, while leaving raw YAML submission unchanged?

src/components/KuadrantTLSCreatePage.tsx
```diff
@@ -271,7 +271,12 @@ const KuadrantTLSCreatePage: React.FC = () => {
                     </HelperText>
                   </FormHelperText>
                 </FormGroup>
-                <GatewaySelect selectedGateway={selectedGateway} onChange={setSelectedGateway} />
+                <GatewaySelect
+                  selectedGateway={selectedGateway}
+                  onChange={setSelectedGateway}
+                  namespace={selectedNamespace}
+                  isDisabled={formDisabled}
```
jasonmadigan: this disables Gateway retargeting when editing a TLSPolicy, but the existing smoke test deliberately retargets it and the API supports that update. DNSPolicy now has the same change. could we retain the new namespace scoping without disabling the DNS/TLS selectors? if immutability is intentional, the contract and existing tests need updating.

e2e/tests/policy-forms.spec.ts
```diff
@@ -714,18 +879,106 @@ test.describe('TokenRateLimitPolicy form', () => {
...
+      applyManifest(httprouteManifest(httproute, namespace));
     });
 
     test.afterEach(async () => {
       deleteNamespace(namespace);
     });
 
+    test(
+      'creates a TokenRateLimitPolicy targeting an HTTPRoute via the form',
+      { tag: '@nightly' },
```
jasonmadigan: this and the PlanPolicy→GRPCRoute test at line 1351 are CRUD tests, but they’re tagged `@nightly`. the repository requires CRUD coverage to be `@smoke`, and the current smoke run excludes both new save paths. could we tag both tests `@smoke`?

e2e/tests/policy-forms.spec.ts
```diff
@@ -714,18 +879,106 @@ test.describe('TokenRateLimitPolicy form', () => {
...
+      - limit: 100
+        window: 1m
+`,
+        );
+
+        await page.getByRole('tab', { name: 'Form' }).click();
+
+        // the unsupported kind must not silently populate a target - Create
+        // stays disabled rather than persisting a reference the CRD rejects
+        await expect(page.getByRole('button', { name: 'Create', exact: true })).toBeDisabled();
```
jasonmadigan: this assertion is already true on the initial empty form, so the test passes even if Monaco never invokes the YAML hydration callback. could we first establish a valid, Create-enabled target, then inject the unsupported YAML and assert that its name/rules hydrated while the target was cleared? that makes the disabled result causally test the new guard.

e2e/tests/policy-forms.spec.ts
```diff
@@ -578,18 +628,133 @@ test.describe('OIDCPolicy form', () => {
...
+        );
+
+        await page.locator('#policy-name').fill(`e2e-oidc-${uid()}`);
+        await page.locator('#client-id').fill('my-client-id');
+        await page.locator('#issuer-url').fill('https://auth.example.com');
+
+        const gatewayOption = page.locator(
+          `#gateway-select option[value="${namespace}/${gateway}"]`,
+        );
+        await expect(gatewayOption).toBeAttached({ timeout: 15_000 });
```
jasonmadigan: this proves the local Gateway is present, but the old cluster-wide selector would also pass. could we assert that the seeded foreign-namespace Gateway is absent, and add focused coverage that `#ALL_NS#` disables the selector and shows the namespace warning?

jasonmadigan (changes requested):

jasonmadigan (approved):

jasonmadigan:
clockit

Issue #776, opened 2026-08-25
Validate the MCP Inspector backend proxy approach


jasonmadigan:
## What

Validate and record the transport architecture for the MCP Inspector in #671. The PoC is complete; production hardening remains part of the Tools work.

Implementation draft: #779.

## Decision

Replace nginx in the Console Plugin image with a small Go server. The same process serves the compiled frontend and a narrow MCP relay on port 9443.

The Kuadrant Operator continues to own the Deployment, Service, serving certificate and `ConsolePlugin` resource. The relay does not run in the operator process. Its source and binary live in `kuadrant-console-plugin`, beside the UI that depends on it.

```text
Browser
  -> /api/proxy/plugin/kuadrant-console-plugin/backend/
       api/mcp/v1/mcpgatewayextensions/{namespace}/{name}
  -> OpenShift Console Bridge
       injects the signed-in user's OpenShift token
  -> Console Plugin Go server
       reads the extension and referenced Gateway as that user
       derives and validates the target
       removes the OpenShift token
       adds the separate MCP bearer token, if supplied
  -> MCP Gateway
```

The browser supplies an `MCPGatewayExtension` namespace and name, never a destination URL.

## Why use a backend

MCP Gateways can use any ingress hostname. A finite `ConsolePlugin.spec.contentSecurityPolicy.connectSrc` list cannot describe every gateway or identity provider in a cluster. Direct browser access also makes gateway CORS part of the feature contract.

The fixed Console proxy keeps browser traffic same-origin. Gateway CORS, dynamic Console CSP entries, custom HTTPRoutes, MCP Gateway PR #1416 and `status.mcpEndpoint` are not prerequisites.

This revisits two earlier designs:

- PR #674 put a proxy in the Developer Portal controller. That would couple the Console feature to an optional, independently versioned component.
- PR #748 used direct browser-to-gateway requests. The dynamic CSP problem makes that a poor fit for Console.

Keeping the relay in the Console Plugin image reuses the existing Service, certificate and image lifecycle without exposing an application endpoint from the privileged operator process.

## First slice

Supported:

- `initialize` and `notifications/initialized`;
- `tools/list`, including refresh;
- `tools/call`;
- JSON and finite request-scoped SSE responses;
- sessions held in browser memory;
- an optional manually entered bearer token.

Not included: OIDC, Prompts, Logs, open notification streams, server-side session storage or a general-purpose URL proxy.

## Target resolution

For each MCP request, the PoC:

[...]

Issue #671, opened 2026-07-17
MCP Inspector — Tools


R-Lawton:
Implement the MCP Inspector page with the Tools tab as shown in the Figma designs. This is the core interactive tool for connecting to MCP gateways and testing tools.

Related architecture validation: #776.

## Plugin wiring

- Add `console.page/route` for MCP Inspector.
- Add the exposed module in `package.json` under `consolePlugin.exposedModules`.
- Run `yarn i18n` for new translatable strings.

## Connection bar

- `Card` with three sections: Gateway | Connection | Status.
- `Select` populated by watching `MCPGatewayExtension` resources with `useK8sWatchResource`.
- A connectable extension has a current `Ready=True` condition. The UI does not depend on `status.mcpEndpoint`.
- Connection status: `Icon` + Connected / No connection.
- Authentication status when a gateway bearer token is in use.
- Request, warning and error counts.

## Empty state

Show “No connection: Connect to a Gateway to view the MCP server tools available” until a gateway is selected.

## Authentication

When the gateway returns `401`, show an “Authentication required” modal with a password input for a bearer token. Keep the token in browser memory only.

The OpenShift user token and MCP gateway token are separate credentials. OIDC is not part of this issue.

## Tools workspace

Left panel:

- Searchable tool selector with an icon-only refresh button.
- Tool name, MCP server name, description and annotation labels.
- Copy action.
- Inputs generated from the tool's JSON schema.
- Optional tool-specific `_meta` key/value rows.
- Validate-only and Run tool actions where supported.

Right panel:

- Console and server-result views.
- Response text and JSON-RPC request/response payloads.
- Aligned status and duration summary.

Prompts and Logs remain future tabs; only Tools is wired up here.

## MCP transport

Browser MCP traffic is same-origin and goes through the OpenShift Console plugin proxy:

```text
Browser
  -> /api/proxy/plugin/kuadrant-console-plugin/backend/
       api/mcp/v1/mcpgatewayextensions/{namespace}/{name}
  -> Console Plugin Go backend
  -> MCP Gateway
```

The Console Plugin image contains a small Go server that serves the compiled frontend and exposes the narrow MCP relay. The Kuadrant Operator deploys this image and reconciles a fixed `ConsolePlugin.spec.proxy` entry with `authorization: UserToken`. The relay does not run inside the operator process.

[...]

PR #779, opened 2026-08-26
Add MCP Inspector Tools and Prompts

### LINKED ISSUE #672: MCP Inspector — Prompts
Add the Prompts tab to the MCP Inspector. Depends on Kuadrant/kuadrant-console-plugin#671 for the page shell, connection bar, and MCP session management.

**Prompts tab:**
- `Alert` (info, inline) — "Generating prompts creates text templates only and does not execute commands"
- **Left panel — "Select a prompt":**
  - `Select` (typeahead) with refresh `Button`
  - Prompt name with sparkle `Icon` + MCP server name
  - Description text
  - Copy `Button` (icon)
  - Dynamic `FormGroup` inputs from prompt arguments (`TextInput`, `Select`)
  - `Button` (primary): "Generate prompt"
  - `Button` (secondary): "Clear fields"
- **Right panel — "Output":**
  - `CodeBlock` showing generated prompt text
  - Token count display

## MCP protocol

| Action | MCP Method | Proxy endpoint |
|---|---|---|
| List prompts | `prompts/list` | `POST /gateways/<ns>/<name>/mcp` |
| Get prompt | `prompts/get` | `POST /gateways/<ns>/<name>/mcp` |

**Designs:** https://www.figma.com/proto/oizjCN6EVed31dW71YXS3K/RHCL-MCP?node-id=332-19491

Static copy:

Image

## Definition of done
- [ ] Unit tests (Jest)
- [ ] E2e tests (Playwright)
- [ ] Documentation updated

jasonmadigan:
## Summary

- Add an MCP Inspector for tools and prompts, with schema-driven inputs, JSON-RPC output, and automatic or explicit selection of stateful `2025-11-25` and stateless `2026-07-28` protocols.
- Serve plugin assets and a same-origin Go backend relay. Gateway targets are resolved using the Console user's Kubernetes permissions; Console and MCP credentials stay separate, tokens stay in memory, and upstream requests have method, size, timeout and redirect restrictions.
- Use the operator-managed backend and Console proxy contract. Local setup includes separate stateful and stateless demo servers, scoped development network access, and browser journeys for both protocols.

Depends on Kuadrant/kuadrant-operator#2206, which reconciles `ConsolePlugin.spec.proxy` (`authorization: UserToken`), runs the plugin image as the backend on 9443 and honours `CONSOLE_PLUGIN_IMAGE_OVERRIDE`.

## Demo

https://github.com/user-attachments/assets/4e59285a-394f-45cb-9dc1-c6d5e62414f2

## Try it locally

The Kuadrant Operator is the product source of truth: it deploys the backend-capable Console Plugin image and reconciles `ConsolePlugin.spec.proxy`. oinc only adapts that contract for its standalone development Console.

The Bridge code path is the production one: `authorization: UserToken` wraps the proxy route in the Console auth middleware, which injects the authenticated user's token and verifies CSRF. With auth disabled the Bridge injects a static token instead, exactly as `start-console.sh` does for every plugin, so the one thing this flow cannot show is per-user RBAC behaviour; check that on a real cluster with a non-admin user before GA.

Needs [oinc v0.4.6](https://github.com/jasonmadigan/oinc/releases/tag/v0.4.6) or newer, Docker or Podman, kubectl, jq and Node.js. Assume sibling checkouts named `kuadrant-console-plugin` (this branch) and `kuadrant-operator` (`feat/mcp-inspector-backend-proxy`).

### 1. Start oinc and frontend hot reload

From this checkout, leave running:

```bash
make oinc
```

On a clean workspace this creates the oinc cluster, installs Kuadrant and Istio, applies the dual-protocol MCP demo, and starts the frontend dev server on port 9001. Current Kuadrant provides the MCP controller and CRDs; do not install a second standalone MCP Helm addon alongside it. Only set `MCP_GATEWAY_ADDON=mcp-gateway` when deliberately pinning a pre-consolidation Kuadrant release that does not supply MCP.

For an existing cluster, `make oinc` skips resource setup. Step 7 installs or refreshes both demo servers without recreating the cluster. Run steps 2 onward in a second terminal, using the `oinc` kubectl context.

### 2. Build and load the backend-capable Console Plugin image

```bash
docker build -t localhost/kuadrant/console-plugin:mcp-inspector-dev .
oinc load-image localhost/kuadrant/console-plugin:mcp-inspector-dev
```

### 3. Build and load the development Kuadrant Operator

```bash
cd ../kuadrant-operator
git checkout feat/mcp-inspector-backend-proxy
docker build --build-arg WITH_EXTENSIONS=false \
  -t localhost/kuadrant/kuadrant-operator:mcp-inspector-dev .
oinc load-image localhost/kuadrant/kuadrant-operator:mcp-inspector-dev
cd -
```

### 4. Point the OLM-managed operator at both local images

Patch the owning CSV, not the generated Deployment. The image override also enables Console Plugin reconciliation in oinc, which has no `ClusterVersion` object. Idempotent, safe to re-run.

```bash
KUADRANT_CSV=$(kubectl get subscription kuadrant-operator \
  --namespace kuadrant-system \
  -o jsonpath='{.status.installedCSV}')

kubectl get csv "${KUADRANT_CSV}" \
  --namespace kuadrant-system \
  -o json |
  jq \
    --arg operator_image 'localhost/kuadrant/kuadrant-operator:mcp-inspector-dev' \
    --arg plugin_image 'localhost/kuadrant/console-plugin:mcp-inspector-dev' \
    '
      .spec.install.spec.deployments[0].spec.template.spec.containers[0].image = $operator_image |
      .spec.install.spec.deployments[0].spec.template.spec.containers[0].imagePullPolicy = "IfNotPresent" |
      .spec.install.spec.deployments[0].spec.template.spec.containers[0].env = (
        (.spec.install.spec.deployments[0].spec.template.spec.containers[0].env // [] |
          map(select(.name != "CONSOLE_PLUGIN_IMAGE_OVERRIDE"))) +
        [{"name":"CONSOLE_PLUGIN_IMAGE_OVERRIDE","value":$plugin_image}]
      )
    ' |
  kubectl replace -f -

kubectl rollout status deployment/kuadrant-operator-controller-manager -n kuadrant-system --timeout=2m
kubectl rollout status deployment/kuadrant-console-plugin -n kuadrant-system --timeout=2m
kubectl get consoleplugin kuadrant-console-plugin -o jsonpath='{.spec.proxy}{"\n"}'
```

### 5. Add oinc-only HTTP routing settings

The oinc demo gateway is plain HTTP and its public sslip hostname resolves to loopback inside the plugin pod. Keep the logical URL and Host header, but dial the in-cluster Gateway Service. The operator retains environment variables it does not own.

```bash
kubectl set env deployment/kuadrant-console-plugin -n kuadrant-system \
  MCP_PROXY_DIAL_ADDRESS=mcp-gateway-istio.gateway-system.svc.cluster.local:80 \
  MCP_PROXY_ALLOW_INSECURE_AUTH=true
kubectl rollout status deployment/kuadrant-console-plugin -n kuadrant-system --timeout=2m
```

### 6. Sync the operator proxy into standalone Console

```bash
make oinc-sync-plugin-proxy
```

This also refreshes the development-only NetworkPolicy for the current Console container IP and checks backend health through Console. Run it again after recreating Console, recreating the cluster, or changing the backend Service.

### 7. Install or refresh both demo servers

Run this for an existing cluster, or re-run it to confirm a fresh setup:

```bash
kubectl config use-context oinc
make oinc-mcp-demo
kubectl --context=oinc get deployments,mcpserverregistrations -n toystore
```

The command applies `scripts/mcp-demo.yaml`: demo namespaces, the MCP Gateway and its ReferenceGrant, MCPGatewayExtension, and both servers with their Services, HTTPRoutes and registrations. It waits for both Deployments to roll out and for the extension and registrations to become Ready.

Expected resources:

- `mcp-test-server` Deployment and `toystore-mcp-server` registration: existing stateful server, prefix `toystore_`.
- `mcp-test-stateless-server` Deployment and `stateless-mcp-server` registration: upstream stateless sample, prefix `stateless_`.

Both Deployments should show `1/1` Ready and both registrations should show `READY=True`.

### 8. Use the Inspector with both protocols

Open <http://localhost:9000/mcp-inspector> and select `mcp-gateway-extension (mcp-gateway-system)`.

| Protocol selector | Tool (`Name=Ada`) | Prompt |
|---|---|---|
| `2025-11-25` | `toystore_greet` → `Hi Ada` | `toystore_greet`, no declared arguments → `Say hi to ` |
| `2026-07-28` | `stateless_hello_world` → `Hello, Ada!` | `stateless_greeting`, `Name=Ada` → `Please greet Ada warmly` |

The new backend uses upstream's `ghcr.io/kuadrant/mcp-gateway/test-stateless-server:latest` sample. The existing `test-server1:latest` deployment and `toystore_` prefix are unchanged.

**Auto prefers the stateless catalog** once both backends are discovered. Select `2025-11-25` explicitly to see the stateful tools. Reconnect after adding a backend if needed.

Stateful tool and prompt journey:

```bash
curl -fsS http://localhost:9000/api/proxy/plugin/kuadrant-console-plugin/backend/healthz

MCP_INSPECTOR_E2E_EXTENSION=mcp-gateway-system/mcp-gateway-extension \
  MCP_INSPECTOR_E2E_PROTOCOL=2025-11-25 \
  npx playwright test --config=e2e/playwright.config.ts \
  e2e/tests/mcp-inspector.spec.ts -g 'connects to a live gateway'
```

Stateless tool and prompt journey:

```bash
MCP_INSPECTOR_E2E_EXTENSION=mcp-gateway-system/mcp-gateway-extension \
  MCP_INSPECTOR_E2E_PROTOCOL=2026-07-28 \
  MCP_INSPECTOR_E2E_TOOL=stateless_hello_world \
  MCP_INSPECTOR_E2E_PROMPT=stateless_greeting \
  MCP_INSPECTOR_E2E_PROMPT_ARGUMENT_LABEL=Name \
  MCP_INSPECTOR_E2E_PROMPT_OUTPUT='Please greet Ada warmly' \
  npx playwright test --config=e2e/playwright.config.ts \
  e2e/tests/mcp-inspector.spec.ts -g 'connects to a live gateway'
```

Repeat the stateless command with `MCP_INSPECTOR_E2E_PROTOCOL=auto` to check negotiation. The test's default protocol is now `2025-11-25` to match its default `toystore_greet` fixture; the UI still defaults to Auto. See [the demo guide](https://github.com/jasonmadigan/kuadrant-console-plugin/blob/feat/mcp-inspector-tools/docs/mcp-inspector.md#demo-servers).

### 9. Optional: exercise the bearer token prompt

The demo gateway is open, so the Inspector connects without a credential. Protect the `mcp` listener to see the `401` flow. Run step 8 first, or delete the policy afterwards: the live Playwright journey has no token step.

```bash
kubectl apply -f - <<'YAML'
apiVersion: v1
kind: Secret
metadata:
  name: mcp-inspector-bearer
  namespace: kuadrant-system
  labels:
    app: kuadrant-mcp-inspector
    authorino.kuadrant.io/managed-by: authorino
stringData:
  api_key: test
---
apiVersion: kuadrant.io/v1
kind: AuthPolicy
metadata:
  name: mcp-inspector-bearer
  namespace: gateway-system
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: mcp-gateway
    sectionName: mcp
  rules:
    authentication:
      bearer-token:
        apiKey:
          selector:
            matchLabels:
              app: kuadrant-mcp-inspector
        credentials:
          authorizationHeader:
            prefix: Bearer
    response:
      unauthenticated:
        code: 401
        headers:
          WWW-Authenticate:
            value: Bearer
        body:
          value: |
            {"error": "Unauthorized", "message": "Authentication required."}
YAML

kubectl wait authpolicy/mcp-inspector-bearer -n gateway-system --for=condition=Enforced --timeout=2m
```

Select protocol `2025-11-25` and the gateway again. The "Authentication required" modal opens; a wrong token shows "Invalid bearer token" inline and keeps the modal open, `test` connects and lists the `toystore_*` tools.

### Rebuild loop

Frontend changes hot reload. For Go backend or operator changes, rebuild and `oinc load-image` the image, then `kubectl rollout restart` the matching Deployment; the CSV references keep the stable dev tags.

## Test evidence

- Dual-protocol demo on oinc: both registrations Ready; repeat setup succeeds; live tools/prompts journeys pass for explicit `2025-11-25`, explicit `2026-07-28`, and Auto → stateless (retries disabled).
- Five demo/network-policy fixture tests pass, plus TypeScript, targeted ESLint, and the spec-map check.
- `go test ./...`
- `yarn lint`
- `yarn test` (MCP Inspector Jest suites: `src/components/mcp/MCPInspectorPage.test.tsx`, `src/utils/mcp/client.test.ts`, `src/utils/mcp/serverNames.test.ts`)
- Live browser journey against the setup above, including the bearer prompt: `initialize=200`, `notifications/initialized=202`, `tools/list`, `tools/call` returning `Hi Ada`, `prompts/list`, `prompts/get` returning `Say hi to `, no gateway CSP report.

Relates to #671. Closes #672. Validation of the backend proxy approach: #776.

## Summary by CodeRabbit

- **New Features**
  - Added an MCP Inspector for browsing gateways, tools, and prompts; submitting inputs; running tools; and reviewing results and telemetry.
  - Added administrator and developer navigation links.
  - Added secure HTTPS proxying with bearer-token authentication and modern and legacy MCP protocol support.
  - Added local development proxy synchronisation and MCP demo setup.
  - Added deployment health checks and private certificate authority support.

- **Documentation**
  - Added MCP Inspector usage, authentication, configuration, and development guidance.

- **Bug Fixes**
  - Improved authentication failures, expired sessions, validation, pagination, unsupported prompt handling, and registration form synchronisation.


src/components/mcp/MCPInspectorPage.test.tsx
```diff
@@ -0,0 +1,434 @@
...
+      sendInitialized: jest.fn().mockResolvedValue(undefined),
+      toolsList: mockToolsList,
+      toolsCallWithDetails: mockToolsCallWithDetails,
+      promptsList: mockPromptsList,
+      promptsGetWithDetails: mockPromptsGetWithDetails,
+    }));
+  });
+
+  it('guides the user to select a gateway before showing inspector tools', () => {
+    render(<MCPInspectorPage />);
```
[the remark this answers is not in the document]
jasonmadigan: Won't fix. That guideline covers manual checks in both Console themes. jsdom computes no theme styles, so a jest render under a dark theme class would assert nothing.

src/components/mcp/MCPInspectorPage.tsx
```diff
@@ -0,0 +1,685 @@
...
+    await client.sendInitialized();
+    const listed = await client.toolsList();
+    const promptState = await listPrompts(client);
+    clientRef.current = client;
+    setSessionId(id);
+    setConnected(true);
+    setTools(listed.tools ?? []);
+    setPrompts(promptState.prompts);
+    setPromptSupport(promptState.support);
+    setPromptError(promptState.error);
```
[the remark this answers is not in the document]
jasonmadigan: Applied in b616814. Connect attempts carry a generation counter; a gateway change or a newer connect invalidates anything still in flight, so a slow gateway can no longer land its session, tools, prompts or 401 challenge under the newly selected one. Covered by a test that resolves the earlier connect late.

src/components/mcp/MCPToolWorkspace.tsx
```diff
@@ -0,0 +1,435 @@
...
+        args[name] = parsed;
+        return;
+      }
+
+      if (type === 'object' || type === 'array') {
+        try {
+          const parsed = JSON.parse(String(value));
+          if (
+            (type === 'array' && !Array.isArray(parsed)) ||
+            (type === 'object' && (Array.isArray(parsed) || typeof parsed !== 'object'))
```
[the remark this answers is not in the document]
jasonmadigan: Applied in b616814. null is now rejected for object and array inputs unless the schema type includes 'null'. Test covers both cases.

src/components/mcp/MCPToolWorkspace.tsx
```diff
@@ -0,0 +1,435 @@
...
+      setFieldErrors((current) => ({ ...current, [name]: '' }));
+      setValidationMessage('');
+    };
+
+    if (propertySchema.enum) {
+      return (
+        <FormSelect
+          id={id}
+          value={String(value)}
+          onChange={(_event, next) => setValue(next)}
```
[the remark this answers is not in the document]
jasonmadigan: Applied in b616814. validate resolves the selected string back to the schema's own enum entry, so a boolean enum sends true rather than "true"; an unmatched value is a validation error. Test covers a boolean enum.

src/utils/mcp/client.ts
```diff
@@ -0,0 +1,334 @@
...
+  async sendInitialized(): Promise<void> {
+    const res = await this.post({ jsonrpc: '2.0', method: 'notifications/initialized' });
+    if (!res.ok) {
+      throw this.httpErrorFor(res.status, 'notifications/initialized failed');
+    }
+  }
+
+  async toolsList(): Promise<ToolsListResult> {
+    return this.call<ToolsListResult>('tools/list', {});
+  }
```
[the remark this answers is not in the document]
jasonmadigan: Applied in b616814. tools/list and prompts/list follow nextCursor and concatenate pages, stopping with an error on a repeated cursor or past 100 pages. Tests cover a two-page list and a cursor that never advances.

jasonmadigan:
Ready

src/components/mcp/MCPInspectorPage.tsx
```diff
@@ -0,0 +1,702 @@
...
+    }
+    setError('');
+    setSessionExpired(false);
+    setCalling(true);
+    setCallExchange(null);
+    try {
+      const exchange = metadata
+        ? await client.toolsCallWithDetails(toolName, args, metadata)
+        : await client.toolsCallWithDetails(toolName, args);
+      setCallExchange(exchange);
```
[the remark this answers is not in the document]
jasonmadigan: Applied in 50e525c. Tool calls, prompt generation and both refreshes now run through one helper that records the connect attempt they started on and drops the result, the failure and the busy flag if the gateway changed meanwhile. Changing gateway also resets every busy flag. Test covers a tool result landing after a switch.

src/components/mcp/MCPToolWorkspace.tsx
```diff
@@ -0,0 +1,452 @@
...
+const initialValues = (tool: MCPTool): FieldValues => {
+  const properties = (tool.inputSchema as JsonSchema | undefined)?.properties ?? {};
+  return Object.entries(properties).reduce<FieldValues>((values, [name, schema]) => {
+    if (schema.default !== undefined) {
+      values[name] =
+        typeof schema.default === 'boolean'
+          ? schema.default
+          : typeof schema.default === 'string'
+          ? schema.default
+          : JSON.stringify(schema.default, null, 2);
```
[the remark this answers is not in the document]
jasonmadigan: Applied in 50e525c, without a separate null control. A null default now means no default: the field starts empty and, being optional, is omitted from the call, which is what the server's default expresses. Nullable object inputs still send an explicit null via the earlier rule. Test covers the omitted field.

src/utils/mcp/client.ts
```diff
@@ -0,0 +1,366 @@
...
+      .split(/\r?\n/)
+      .filter((line) => line.startsWith('data:'))
+      .map((line) => line.slice(5).trim());
+    if (dataLines.length === 0) {
+      continue;
+    }
+    try {
+      const parsed = JSON.parse(dataLines.join('\n')) as JsonRpcResponse<T>;
+      if (parsed && parsed.jsonrpc === '2.0') {
+        return parsed;
```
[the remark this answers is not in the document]
jasonmadigan: Applied in 50e525c. parseRpcMessage takes the request id and returns only the envelope with that id carrying result or error, so notifications on a request-scoped stream are skipped; an error envelope with a null id is accepted as the server rejecting our request. Same check on plain JSON bodies. Tests cover a notification followed by the response and a stream that never answers.

Anton-Fil:
pu pu pu, not rebase ?

R-Lawton:
👀

cmd/plugin-server/main.go
```diff
@@ -0,0 +1,486 @@
...
+}
+
+func deriveMCPEndpoint(extension *mcpGatewayExtension, targetGateway *gateway) (string, error) {
+	sectionName := extension.Spec.TargetRef.SectionName
+	for _, listener := range targetGateway.Spec.Listeners {
+		if listener.Name != sectionName {
+			continue
+		}
+
+		host := extension.Spec.PublicHost
```
R-Lawton, line 436: Using spec.publicHost as the backend request target means a user who can create or update a Ready MCPGatewayExtension can cause the plugin pod to send requests to arbitrary internal or external hosts. This creates an SSRF and network-egress risk.

cmd/plugin-server/main.go
```diff
@@ -0,0 +1,486 @@
...
+			roots = x509.NewCertPool()
+		}
+		if !roots.AppendCertsFromPEM(caPEM) {
+			return nil, errors.New("Kubernetes CA file contains no certificates")
+		}
+		kubernetesTLS.RootCAs = roots
+	}
+
+	upstreamTransport := http.DefaultTransport.(*http.Transport).Clone()
+	upstreamTransport.TLSClientConfig = &tls.Config{
```
R-Lawton, line 192: HTTPS requests use only the image's system CA roots. If the Gateway uses a cluster or private CA, the request fails with x509: certificate signed by unknown authority, and there is no upstream CA configuration path.

src/components/mcp/MCPInspectorPage.tsx
```diff
@@ -0,0 +1,755 @@
...
+      const result = await operation(client);
+      if (isCurrent()) {
+        onResult(result);
+      }
+    } catch (err) {
+      if (!isCurrent()) {
+        return;
+      }
+      recordFailure(err);
+      if (err instanceof MCPSessionExpiredError) {
```
R-Lawton, line 355: Session expiry only sets sessionExpired while leaving the expired client active. The alert says to reconnect, but there is no reconnect action, and selecting the current gateway again does not trigger a new connection.

src/components/mcp/MCPToolWorkspace.tsx
```diff
@@ -0,0 +1,453 @@
...
+    // a null default means no value, the same as no default
+    if (schema.default !== undefined && schema.default !== null) {
+      values[name] =
+        typeof schema.default === 'boolean'
+          ? schema.default
+          : typeof schema.default === 'string'
+          ? schema.default
+          : JSON.stringify(schema.default, null, 2);
+    } else {
+      values[name] = schemaType(schema) === 'boolean' ? false : '';
```
R-Lawton: Optional boolean properties without schema defaults are initialized to false. This means every call sends the property even when the user never supplied it, which can override the server's default or unset semantics.

src/components/mcp/MCPInspectorPage.tsx
```diff
@@ -0,0 +1,755 @@
...
+  `/api/proxy/plugin/kuadrant-console-plugin/backend/api/mcp/v1/mcpgatewayextensions/${encodeURIComponent(
+    ext.metadata?.namespace ?? '',
+  )}/${encodeURIComponent(ext.metadata?.name ?? '')}`;
+
+const MCPInspectorPage: React.FC = () => {
+  const { t } = useTranslation('plugin__kuadrant-console-plugin');
+  const [activeNamespace] = useActiveNamespace();
+  const resolvedNamespace = activeNamespace === ALL_NS ? undefined : activeNamespace;
+
+  const [extensions, extensionsLoaded] = useK8sWatchResource<MCPGatewayExtension[]>({
```
R-Lawton: The extension and registration watch errors are ignored. If either list fails because of RBAC or an API/watch error, the selector can remain disabled as if it were still loading, without explaining the failure.

src/components/mcp/MCPInspectorPage.tsx
```diff
@@ -0,0 +1,755 @@
...
+        onResult(result);
+      }
+    } catch (err) {
+      if (!isCurrent()) {
+        return;
+      }
+      recordFailure(err);
+      if (err instanceof MCPSessionExpiredError) {
+        setSessionExpired(true);
+      } else {
```
R-Lawton, line 363: MCPUnauthorizedError from an active tool or prompt call is handled as a generic error. An expired or revoked bearer token does not reopen the authentication dialog, so the user cannot recover without changing the gateway or protocol.

e2e/tests/mcp-inspector.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    await expect(page.getByRole('heading', { name: 'MCP Inspector' })).toBeVisible({
+      timeout: 15_000,
+    });
+    await expect(page.getByLabel('Select an MCP gateway extension')).toBeVisible();
+    await expect(page.getByLabel('MCP protocol', { exact: true })).toHaveValue('auto');
+    await expect(page.getByRole('heading', { name: 'No connection' })).toBeVisible();
+  });
+
+  test('connects to a live gateway and runs a tool', { tag: '@nightly' }, async ({ page }) => {
+    test.skip(
```
R-Lawton, line 43: The live inspector journey is skipped unless MCP_INSPECTOR_E2E_EXTENSION is set. In normal CI this only verifies that the page opens; proxy, authentication, protocol selection, tool/prompt execution, and output handling are not exercised.

cmd/plugin-server/main.go
jasonmadigan: Addressed in 3f4c56a. MCP_PROXY_CA_FILE now appends a PEM bundle to the upstream client's system trust roots, separately from Kubernetes API trust. Helm can mount ca-bundle.crt from plugin.mcpProxy.caConfigMapName. Invalid bundles fail startup, and a Go test covers a successful private-CA HTTPS connection.

src/components/mcp/MCPInspectorPage.tsx
jasonmadigan: Fixed in 2e30b4f. Session expiry now invalidates the client and pending operations. The alert offers Reconnect for the same gateway and protocol, retaining the bearer token. Regression coverage checks recovery and rejects stale completions.

src/components/mcp/MCPToolWorkspace.tsx
jasonmadigan: Fixed in 2e30b4f. Optional booleans without defaults start as Not set and are omitted from arguments. Users can select True or False and return to Not set; explicit false and schema defaults are preserved. Tests cover these transitions.

src/components/mcp/MCPInspectorPage.tsx
jasonmadigan: Addressed in 2e30b4f. Extension watch failures now display an error instead of appearing to load indefinitely, and recover when the watch succeeds. Registration failures show a separate warning because that lookup only enriches server names; connecting and running tools remain available. Both paths have regression coverage.

src/components/mcp/MCPInspectorPage.tsx
jasonmadigan: Fixed in 2e30b4f. A 401 during tool, prompt, or catalog operations invalidates the session, clears the rejected token, and reopens authentication. The same applies when a retained token is rejected during protocol reconnect. Regression tests and the local live bearer rejection/retry journey pass.

e2e/tests/mcp-inspector.spec.ts
jasonmadigan: Addressed in 83cc824. Smoke and full CI now provision the live fixture and run a required Inspector journey step covering legacy, stateless, auto, and bearer rejection/retry. Missing fixture configuration fails that step rather than skipping it. All four journeys passed locally with retries disabled; standalone local runs remain opt-in.

cmd/plugin-server/main.go
```diff
@@ -0,0 +1,486 @@
...
+		s.logger.Warn("MCP upstream request failed", "namespace", namespace, "name", name, "error", err)
+		writeJSONError(writer, http.StatusBadGateway, "MCP gateway request failed")
+		return
+	}
+	defer upstreamResponse.Body.Close()
+	for _, header := range []string{"Content-Type", "Mcp-Session-Id", "MCP-Protocol-Version"} {
+		copyResponseHeader(upstreamResponse.Header, writer.Header(), header)
+	}
+	writer.WriteHeader(upstreamResponse.StatusCode)
+	_, _ = io.Copy(writer, upstreamResponse.Body)
```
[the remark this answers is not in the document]
jasonmadigan: Keeping the current behavior for this Inspector. Its client awaits the complete response with res.text(), including finite request-scoped SSE responses; subscriptions and long-lived streaming are outside the documented scope. Flushing the proxy alone would not provide incremental client handling. The total upstream deadline is intentional and configurable through MCP_PROXY_REQUEST_TIMEOUT.

src/utils/mcp/client.ts
```diff
@@ -0,0 +1,535 @@
...
+function throwOnRpcError(message: JsonRpcResponse<unknown>): void {
+  if (message.error) {
+    throw new MCPRpcError(message.error);
+  }
+  const result = message.result as { resultType?: string } | undefined;
+  if (
+    result?.resultType !== undefined &&
+    !['complete', 'input_required'].includes(result.resultType)
+  ) {
+    throw new Error(`Unsupported MCP result type: ${result.resultType}`);
```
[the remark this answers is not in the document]
jasonmadigan: Keeping compatibility with responses that omit resultType. Although servers are required to emit it, the specification explicitly requires clients to treat an absent resultType as complete for backward compatibility: https://modelcontextprotocol.io/specification/2026-07-28/basic#resulttype. The client already rejects unrecognized values when the field is present.

scripts/cluster-setup.sh
```diff
@@ -96,8 +96,6 @@ spec:
         from: All
 EOF
 
-log "creating demo MCP resources..."
-kubectl create namespace toystore 2>/dev/null || true
-kubectl apply -f "${REPO_DIR}/scripts/mcp-demo.yaml"
+bash "${SCRIPT_DIR}/setup-mcp-demo.sh"
```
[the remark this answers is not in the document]
jasonmadigan: The pinned oinc v0.4.6 already selects the created cluster: its kubeconfig update assigns existing.CurrentContext = clusterName before writing the config. The create path calls that update, so an additional context-switch command is redundant. Source: https://github.com/jasonmadigan/oinc/blob/v0.4.6/pkg/kubeconfig/kubeconfig.go#L53

cmd/plugin-server/main.go
jasonmadigan: The intended trust model is that Gateways and MCPGatewayExtensions are managed by trusted administrators. Their configuration therefore authorizes the proxy destination; Kubernetes read access and Ready status are not being treated as independent network approval.

In 4f640f4, MCP_PROXY_ALLOWED_ORIGINS is optional: an unset or empty list uses the resource-derived destination, while a configured list restricts it to exact origins. This avoids requiring admins to configure each gateway twice. The docs state the trust assumption and the DNS/egress limits; tests cover the default behavior, matching origins, and rejection without an upstream request for non-matching origins. CI fixture setup no longer requires the setting.

jasonmadigan:
@coderabbitai resume

src/components/mcp/MCPToolWorkspace.tsx
```diff
@@ -0,0 +1,495 @@
...
+            value={String(value)}
+            onChange={(_event, next) => setValue(next === '' ? '' : next === 'true')}
+            aria-label={label}
+          >
+            <FormSelectOption value="" label={t('Not set')} />
+            <FormSelectOption value="true" label={t('True')} />
+            <FormSelectOption value="false" label={t('False')} />
+          </FormSelect>
+        );
+      }
```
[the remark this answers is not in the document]
jasonmadigan: Fixed in 6b484d3. Descriptions are now shown for optional boolean and boolean-enum selects; only the required non-enum checkbox suppresses the helper because its label already contains the description. Regression coverage checks all four variants.

jasonmadigan:
@coderabbitai review

jasonmadigan:
Squashed the feature into three signed-off commits: backend relay, Inspector UI/client, and test/development setup.

jasonmadigan:
e2e tests failing likely because of:

https://github.com/Kuadrant/mcp-gateway/pull/1501 https://github.com/Kuadrant/kuadrant-operator/pull/2255 - hopefully fixed soon.

R-Lawton:
Few things i noticed:
1. I think we should add the proctool dropdown to the right of the extension drop down to keep it more in line with figma and to remove the large empty speace under the conection and status 
Screenshot 2026-09-14 at 14 33 48

2. When selecting a tool or prompt you cant click the search bar and get the dropdown you have to click the small arrow to the right wdyt about having the search bar in the dropdown itself ?
Screenshot 2026-09-14 at 14 34 19

3. The words in the left hand output are a bit squashed in some views then spread out in others would be good to get that uniform 

Screenshot 2026-09-14 at 14 33 55

Screenshot 2026-09-14 at 14 35 12

R-Lawton (approved):
Nice to get all this over the line ✨ . Verified it all works as expected locally with both unauthenticated and authenticated.

PR #808, opened 2026-08-31
feat: add external MCP server registration wizard

### LINKED ISSUE #773: Add external MCP server registration
We currently have the ability to register internal MCP server. This issue is to add a form view for external MCP Server registration

Figma Designs: https://www.figma.com/design/oizjCN6EVed31dW71YXS3K/RHCL-MCP?node-id=0-1&p=f&m=draw

emmaaroche:
## Description

Adds an "External" option to the Register MCP Server dropdown on the MCP overview page, opening a 6-step wizard to register an MCP server running outside the cluster. Closes #773.

## Changes

- New `MCPExternalRegistrationWizard` (6 steps): ServiceEntry, DestinationRule, HTTP route, credentials, MCP server registration, verify
- Step components with two-way Form/YAML sync (ServiceEntry, DestinationRule, Credential, RegisterServer)
- Form field components: `ServiceEntryFormFields`, `DestinationRuleFormFields`, `CredentialFormFields`
- Build helpers in `mcpResourceUtils.ts` for ServiceEntry, DestinationRule, credential Secret, MCPServerRegistration
- Sequential resource creation + rollback + readiness watch via shared `MCPVerifyStep`
- `getModelFromResource`: fix core-group ("v1") apiVersion handling so core resources (Secret) resolve correctly
- RBAC: gate External dropdown item on create perms for MCPServerRegistration, ServiceEntry, DestinationRule, Secret
- `resources.ts`: add ServiceEntry / DestinationRule registry entries

## Testing

- New e2e spec `mcp-external-wizard.spec.ts` (6 tests): full flow creates+verifies all resources, tab sync, validation gating, cancel, back
- Updated `mcp-overview.spec.ts` external-option test
- Unit suite 356/356, tsc clean, lint clean, `yarn check:spec-map` OK

## Summary by CodeRabbit

- **New Features**
  - Added external MCP server registration through a guided wizard.
  - Configure ServiceEntry, DestinationRule, credentials, HTTPRoute routing, and MCP server settings using forms or YAML.
  - Added validation, permission checks, resource status handling, cancellation, and backward navigation.
  - External registration is now available when the required permissions are granted.
  - HTTPRoute creation and editing now use a clearer, compact rules table.

- **Bug Fixes**
  - Improved Kubernetes API version handling for core resources.
  - Strengthened validation for hosts, ports, credentials, and routing references.

- **Tests**
  - Added comprehensive coverage for external MCP registration and related resource workflows.


src/components/httproute/HTTPRouteCreatePage.tsx
```diff
@@ -64,6 +62,60 @@ interface HTTPRouteCreatePageProps {
...
+};
+
+// Infer a rule's type from its matches when it wasn't created through the "Rule type"
+// picker (e.g. parsed from YAML or an existing resource being edited).
+const inferRuleType = (matches: HTTPRouteMatch[]): RuleType => {
+  const match = matches?.[0];
+  if (!match) return 'path';
+  if (match.headers?.some((h) => h.name)) return 'header';
+  if (match.queryParams?.some((q) => q.name)) return 'query';
+  if (match.method && !match.pathValue) return 'method';
```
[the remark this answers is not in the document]
emmaaroche: Acknowledged, but not changing this in this PR. A correct fix must preserve path-presence through `parseMatchesFromYAML` and thread it through `inferRuleType`, `validateMatchesInRule` and `generateMatchesForYAML`. Method-match handling is already partly incomplete on `main` (`seedMatchForRuleType('method')` yields `pathType: ''`, which `validateMatchesInRule` rejects), so this is out of scope for the external MCP wizard change.

src/components/mcp/ServiceEntryFormFields.tsx
```diff
@@ -0,0 +1,437 @@
...
+    setErrors((prev) => ({ ...prev, hosts: validateHosts(formState.hosts) }));
+  };
+
+  const handlePortBlur = () => {
+    setTouched((prev) => ({ ...prev, port: true }));
+    setErrors((prev) => ({ ...prev, port: validateServicePort(formState.port) }));
+  };
+
+  return (
+    <Form>
```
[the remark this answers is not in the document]
emmaaroche: These form components render inside `Modal > Wizard > WizardStep`, not as top-level console pages. The `PageSection` guideline applies to route-registered pages; the repo's own `MCPSetupWizard` puts `PageSection` only on the wizard container and keeps step/form bodies as fragments. Wrapping these would nest `PageSection` inside a modal, so leaving as-is.

src/components/mcp/steps/ServiceEntryStep.tsx
```diff
@@ -0,0 +1,101 @@
...
+        protocol: ports?.length ? ports[0].protocol : 'HTTPS',
+        location: typeof spec?.location === 'string' ? spec.location : 'MESH_EXTERNAL',
+        resolution: typeof spec?.resolution === 'string' ? spec.resolution : 'DNS',
+      });
+    } catch {
+      // Invalid YAML — don't update form state
+    }
+  };
+
+  return (
```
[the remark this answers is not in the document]
emmaaroche: Same rationale as the FormFields thread: these steps render inside `Modal > Wizard > WizardStep`. `MCPSetupWizard` applies `PageSection` at the wizard container only and step bodies stay fragments. Wrapping each step in `PageSection` would nest it inside a modal, so leaving as-is.

src/components/httproute/HTTPRouteCreatePage.tsx
```diff
@@ -336,6 +404,8 @@ const HTTPRouteCreatePage: React.FC<HTTPRouteCreatePageProps> = ({ onFormChange
   const formValidation = () => {
     const hasValidParentRef = parentRefs.some((ref) => ref.gatewayName);
 
+    // Gateway API requires spec.rules to have at least one item (minItems=1),
+    // so an HTTPRoute with zero rules is rejected by the API server.
     const hasValidRules =
       rules.length > 0 &&
```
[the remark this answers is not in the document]
emmaaroche: Fixed — alert now reads 'Add at least one rule before creating an HTTPRoute.'

Anton-Fil (commented):
I found four issues that should be addressed before merge. The details are attached inline below.

src/components/mcp/MCPExternalRegistrationWizard.tsx
```diff
@@ -0,0 +1,441 @@
...
+      label: t('Create credential Secret'),
+      resource: buildCredentialSecret(formState.credential, formState.credential.namespace),
+      successMessage: t('Credential Secret created successfully'),
+    });
+
+    items.push({
+      type: 'create' as const,
+      id: 'create-server',
+      label: t('Create MCPServerRegistration'),
+      resource: buildMCPServerRegistration(
```
Anton-Fil, line 178: **Blocking — wire the created Secret into MCPServerRegistration**

This builder creates the registration without `spec.credentialRef`, even though the wizard creates a credential Secret immediately before it. The broker therefore has no way to know which Secret to use for upstream discovery/session management, so the entered token is effectively unused.

Please pass the credential Secret name into this builder and emit:

```yaml
credentialRef:
  name: <secret-name>
  key: token
```

The Secret and registration should be constrained to the same namespace (the reference has no namespace field). Please also extend the e2e assertion to verify `spec.credentialRef.name` and `spec.credentialRef.key`, not only that the Secret exists.

src/components/mcp/MCPExternalRegistrationWizard.tsx
```diff
@@ -0,0 +1,441 @@
...
+        successMessage: t('DestinationRule created successfully'),
+      },
+    ];
+
+    if (routeMode === 'new' && httpRouteResource) {
+      items.push({
+        type: 'create' as const,
+        id: 'create-route',
+        label: t('Create HTTPRoute'),
+        resource: httpRouteResource,
```
Anton-Fil: **Blocking — connect a newly created HTTPRoute to the ServiceEntry**

The new-route form serializes a normal Kubernetes Service backend, while this wizard creates an Istio `ServiceEntry`, not a Kubernetes `Service`. This resource is passed through unchanged here, so the created route is never wired to the external ServiceEntry host. All resources can be created successfully while requests still have no valid path to the external MCP server.

For the Istio external-service flow, generate the implementation-specific backend reference (for example `group: networking.istio.io`, `kind: Hostname`, `name: <external-host>`) or explicitly require an already-configured HTTPRoute. Please add an e2e assertion for the complete `backendRefs` payload.

src/components/mcp/MCPExternalRegistrationWizard.tsx
```diff
@@ -0,0 +1,441 @@
...
+    groupVersionKind: RESOURCES.HTTPRoute.gvk,
+    isList: true,
+    namespace: selectedNamespace,
+  });
+
+  const isHTTPRouteStepValid =
+    (routeMode === 'existing' && !!selectedExistingRouteName) ||
+    (routeMode === 'new' && httpRouteValid && !!httpRouteResource);
+
+  const routeName =
```
Anton-Fil, line 119: **Blocking — preserve the selected HTTPRoute namespace**

Only the selected route name is retained here. If the user chooses an HTTPRoute in namespace `team-a` but creates the MCPServerRegistration in `team-b`, the generated `targetRef` omits `namespace` and resolves to `team-b/github-route` instead of the route the user selected.

Please keep the selected route as `{ name, namespace }` (or keep the full resource), pass the namespace through to `buildMCPServerRegistration`, and add a cross-namespace test. Alternatively, prevent selecting a registration namespace different from the route namespace.

src/components/mcp/MCPExternalRegistrationWizard.tsx
```diff
@@ -0,0 +1,441 @@
...
+    <Modal
+      isOpen={isOpen}
+      onClose={handleClose}
+      variant="large"
+      aria-labelledby="mcp-external-wizard-modal-title"
+    >
+      <ModalHeader
+        labelId="mcp-external-wizard-modal-title"
+        title={t('Setup external MCP server')}
+        description={t('Create gateway by filling out the form or creating the YAML')}
```
Anton-Fil: **Minor — update the modal description**

This copy says `Create gateway`, but this wizard creates resources for registering an external MCP server. That wording can make users think they are creating a Gateway. Please replace it with copy describing the actual flow, for example `Create the resources needed to register an external MCP server`.

Anton-Fil (dismissed):
Image
duplication functionality. Query param and header already exist in http route create wizard

src/components/mcp/mcpResourceUtils.ts
```diff
@@ -144,6 +181,160 @@ export const isMCPServerRegistrationValid = (formState: MCPServerFormState): boo
...
+export const buildCredentialSecret = (
+  formState: CredentialFormState,
+  namespace: string,
+  originalMetadata?: Secret['metadata'] | null,
+): Secret => ({
+  apiVersion: 'v1',
+  kind: 'Secret',
+  metadata: originalMetadata
+    ? { ...originalMetadata, name: formState.credentialName }
+    : { name: formState.credentialName, namespace: formState.namespace || namespace },
```
jasonmadigan: The controller rejects credential Secrets without `mcp.kuadrant.io/secret: "true"` (mcpserverregistration_controller.go:530), so registrations from this wizard go Ready=False and nothing rolls back. Could we add the label here? The e2e also accepts `Error creating resources` as a pass at spec:183 and never asserts Ready, which is why this slipped through; should we assert Ready there?

e2e/tests/mcp-external-wizard.spec.ts
```diff
@@ -0,0 +1,369 @@
...
+): Promise<void> {
+  const wizard = page.locator('.kuadrant-mcp-wizard');
+  await wizard.locator('[data-test="mcp-registration-name"]').fill(opts.name);
+  await wizard.locator('[data-test="mcp-registration-prefix"]').fill(opts.prefix);
+}
+
+async function clickNext(page: Page): Promise<void> {
+  const next = page
+    .locator('.kuadrant-mcp-wizard')
+    .getByRole('button', { name: 'Next', exact: true });
```
jasonmadigan: `clickNext` matches `Next` exactly, but step 5 is now `Save and continue`, so both `@smoke` full-flow tests time out at :177 and :286 (ran locally against a cluster).

e2e/tests/mcp-external-wizard.spec.ts
```diff
@@ -0,0 +1,369 @@
...
+        await clickNext(page);
+
+        await expect(page.getByText('MCPServerRegistration created successfully')).toBeVisible({
+          timeout: 30_000,
+        });
+
+        const route = JSON.parse(
+          kubectl(['get', 'httproute', routeName, '-n', TEST_NAMESPACE, '-o', 'json']),
+        );
+        expect(route.spec.rules[0].backendRefs).toEqual([
```
jasonmadigan: `toEqual` can't match here: the builder emits `port: 443` and the API server defaults `weight: 1`. `toMatchObject` or per-field asserts like :193-236?

src/components/mcp/MCPRegistrationWizard.tsx
```diff
@@ -291,7 +291,7 @@ const MCPRegistrationWizard: React.FC<MCPRegistrationWizardProps> = ({ isOpen, o
               id="register-server"
               isDisabled={!isStep1Valid}
               footer={{
-                nextButtonText: t('Next'),
+                nextButtonText: t('Save and continue'),
```
jasonmadigan: Renaming this and the `watchLabel` at :327 breaks the untouched `mcp-wizard.spec.ts` (`Next` at :175/:277/:388, `MCP server is ready` at :180/:282). 4/8 fail locally, 2 of them `@smoke`.

src/components/mcp/steps/RegisterServerStep.tsx
```diff
@@ -11,119 +11,119 @@ interface RegisterServerStepProps {
...
+        <Tab eventKey="form" title={<TabTitleText>{t('Form')}</TabTitleText>} />
+        <Tab eventKey="yaml" title={<TabTitleText>{t('YAML')}</TabTitleText>} />
+      </Tabs>
+
+      {activeTab === 'form' ? (
         <MCPServerRegistrationFormFields
           formState={formState}
           onChange={handleChange}
           httpRouteNames={routeName ? [routeName] : []}
+          showNamespaceField={false}
```
jasonmadigan: This is unconditional, but the internal wizard's existing-route path only sets namespace through this field and validation still requires it, so `Save and continue` never enables there. Maybe `!credentialNamespace`?

e2e/tests/httproute-crud.spec.ts
```diff
@@ -100,8 +100,7 @@ async function addRuleViaWizard(
 ): Promise<void> {
   const { pathValue, serviceName, servicePort = '8080', isEdit = false, method = 'GET' } = opts;
 
-  // Step 1: Matches — add a match
-  await page.getByRole('button', { name: 'Add match' }).click();
+  // Step 1: Matches — "Add rule" pre-seeds one match, so fill it directly
```
jasonmadigan, line 103: The pre-seeded match was removed again in this commit (`matches: []` at HTTPRouteCreatePage.tsx:366-376), so `#path-type-0` never exists. Should we restore the `Add match` click? Breaks 2 `@smoke` + 1 `@nightly`.

src/components/mcp/MCPExternalRegistrationWizard.tsx
```diff
@@ -0,0 +1,465 @@
...
+        label: t('Create DestinationRule'),
+        resource: buildDestinationRule(
+          formState.destinationRule,
+          formState.destinationRule.namespace,
+        ),
+        successMessage: t('DestinationRule created successfully'),
+      },
+    ];
+
+    if (routeMode === 'new' && httpRouteResource) {
```
jasonmadigan, line 151: Only new-route mode gets wired to the ServiceEntry host. Existing-route mode accepts any route, and the broker then sends the token as `Authorization` (broker/upstream/mcp.go:105) to whatever that route's backend is, e.g. an in-cluster Service. Could we filter to `networking.istio.io/Hostname` backends, or rewire the selected route too?

src/components/mcp/MCPOverviewPage.tsx
```diff
@@ -228,6 +246,16 @@ const MCPOverviewPage: React.FC = () => {
     create: userRBAC[`${getResourceNameFromKind('MCPServerRegistration')}-create`],
   };
 
+  // The external wizard also creates a ServiceEntry, DestinationRule and Secret,
+  // so gate its dropdown item on create permission for all of them.
+  const externalRBAC = {
+    create:
+      serverRBAC.create &&
```
jasonmadigan, line 253: Gate covers SE/DR/Secret/Registration but not HTTPRoute create (new-route mode) or delete (rollback). Rollback failures are only `console.error`'d in MCPVerifyStep:174, which can leave the credential Secret behind with no UI signal. Worth adding both?

src/components/mcp/MCPExternalRegistrationWizard.tsx
```diff
@@ -0,0 +1,465 @@
...
+        ),
+        successMessage: t('HTTPRoute created successfully'),
+      });
+    }
+
+    items.push({
+      type: 'create' as const,
+      id: 'create-credential',
+      label: t('Create credential Secret'),
+      resource: buildCredentialSecret(formState.credential, formState.credential.namespace),
```
jasonmadigan, line 169: MCPVerifyStep treats 409 as `Already exists` success. For a credential Secret that silently drops the token just typed and binds the registration to whatever Secret already had that name. Should we fail or confirm on 409 for this one?

src/components/mcp/steps/ServiceEntryStep.tsx
```diff
@@ -0,0 +1,101 @@
...
+    try {
+      const parsed = yaml.load(yamlInput) as Record<string, unknown>;
+      if (!parsed || typeof parsed !== 'object') return;
+
+      const metadata = parsed.metadata as Record<string, string> | undefined;
+      const spec = parsed.spec as Record<string, unknown> | undefined;
+      const ports = spec?.ports as Array<{ number: number; protocol: string }> | undefined;
+      const hosts = spec?.hosts as string[] | undefined;
+
+      onChange({
```
jasonmadigan: YAML edits update state but never revalidate; `isServiceEntryValid` / `isDestinationRuleValid` / `isCredentialValid` have no callers, so a name blanked in YAML keeps Next enabled. Same in the other steps. Could we call the validator from `handleYamlChange`?

src/components/mcp/steps/RegisterServerStep.tsx
```diff
@@ -11,119 +11,119 @@ interface RegisterServerStepProps {
...
+      (credentialNamespace && formState.namespace !== credentialNamespace)
+    ) {
+      onChange({
+        ...formState,
+        ...(routeName ? { targetHTTPRouteName: routeName } : {}),
+        ...(credentialNamespace ? { namespace: credentialNamespace } : {}),
+      });
     }
-  }, [routeName]); // eslint-disable-line -- only resync when routeName changes; re-running on formState/onChange would fight typing in other fields
+  }, [routeName, credentialNamespace]); // eslint-disable-line -- only resync wizard-owned references
```
jasonmadigan, line 46: PF only mounts the active step, so changing the credential namespace on step 4 and jumping to step 6 via the nav leaves the registration in the old namespace, where the controller can't find the Secret. Deriving the namespace at build time (wizard:179) would avoid it.

src/components/mcp/MCPExternalRegistrationWizard.tsx
```diff
@@ -0,0 +1,465 @@
...
+            <WizardStep
+              name={t('Create Destination Rule')}
+              id="create-destination-rule"
+              isDisabled={!isServiceEntryValid}
+              footer={{
+                nextButtonText: t('Next'),
+                isNextDisabled: !isDestinationRuleValid,
+              }}
+            >
+              <DestinationRuleStep
```
jasonmadigan, line 287: DR host and namespace are retyped with no prefill or check against the ServiceEntry; a mismatch means the DR is ignored and no TLS origination. `DISABLE` is also offered while the token is mandatory. Could we prefill from `parseServiceEntryHosts(...)[0]` and the SE namespace?

jasonmadigan (changes requested):
Thanks for the credentialRef, backendRef and targetRef namespace fixes. A few blockers left, mainly the credential Secret label and some e2e regressions from the button renames. Happy to re-review once addressed.

jasonmadigan:
:eyes:

emmaaroche:
Fixing tests

src/components/mcp/steps/RegisterServerStep.tsx
```diff
@@ -1,129 +1,131 @@
...
+        <Tab eventKey="form" title={<TabTitleText>{t('Form')}</TabTitleText>} />
+        <Tab eventKey="yaml" title={<TabTitleText>{t('YAML')}</TabTitleText>} />
+      </Tabs>
+
+      {activeTab === 'form' ? (
         <MCPServerRegistrationFormFields
           formState={formState}
           onChange={handleChange}
           httpRouteNames={routeName ? [routeName] : []}
+          showNamespaceField={false}
```
jasonmadigan: This is now unconditional, and `MCPRegistrationWizard` passes no `credentialNamespace`. It only sets `formState.server.namespace` inside the `routeMode === 'new'` branch (MCPRegistrationWizard.tsx:275-279), so in the default `existing` mode the namespace stays `''` from `initialServerFormState` (types.ts:150), `validateRegistrationNamespace` fails, and step 2's `Next` never enables. There's no field left to fix it.

Also removes `#server-namespace`, which `fillStep2` (mcp-wizard.spec.ts:101) and `step 2 namespace can be selected` (:189) both need. 2 `@smoke` + 2 `@nightly`.

```suggestion
          showNamespaceField={!credentialNamespace}
```

e2e/tests/httproute-crud.spec.ts
jasonmadigan, line 103: Right fix here, but `Add match` only renders when `matches.length === 0` (HTTPRouteRuleWizard.tsx:299-311); once seeded it's a `Tabs onAdd` with no matching accessible name. :353, :424 and :598 still click it and will time out. 3 `@nightly`. Same treatment as here, the seeded match is already at index 0.

src/components/mcp/MCPExternalRegistrationWizard.tsx
```diff
@@ -0,0 +1,489 @@
...
+      allowAlreadyExists: false,
+    });
+
+    items.push({
+      type: 'create' as const,
+      id: 'create-server',
+      label: t('Create MCPServerRegistration'),
+      resource: buildMCPServerRegistration(
+        formState.server,
+        formState.server.namespace,
```
jasonmadigan: `formState.server.namespace` is only kept in sync by `RegisterServerStep`'s effect, and PF mounts just the active step. Visit step 5, go back to step 4, change the credential namespace, then click step 6 in the nav: step 5 never remounts, so the registration lands in the old namespace while the Secret lands in the new one. Controller can't resolve the Secret, `Ready=False`, rollback fires.

Reading the credential namespace directly avoids the resync entirely.

```suggestion
        formState.credential.namespace,
```

src/components/mcp/MCPExternalRegistrationWizard.tsx
```diff
@@ -0,0 +1,489 @@
...
+    t,
+  ]);
+
+  const verifyWatchResource = React.useMemo<WatchResourceConfig>(
+    () => ({
+      gvk: RESOURCES.MCPServerRegistration.gvk,
+      name: formState.server.registrationName,
+      namespace: formState.server.namespace,
+    }),
+    [formState.server.registrationName, formState.server.namespace],
```
jasonmadigan: Same here, otherwise the watch points at the stale namespace.

```suggestion
      gvk: RESOURCES.MCPServerRegistration.gvk,
      name: formState.server.registrationName,
      namespace: formState.credential.namespace,
    }),
    [formState.server.registrationName, formState.credential.namespace],
```

src/components/mcp/MCPExternalRegistrationWizard.tsx
```diff
@@ -0,0 +1,489 @@
...
+              }
+              footer={{
+                nextButtonText: t('Finish'),
+                isNextDisabled: !resourcesCreated,
+              }}
+            >
+              <MCPVerifyStep
+                items={verifyItems}
+                watchResource={verifyWatchResource}
+                selectedNamespace={formState.server.namespace}
```
jasonmadigan: And here.

```suggestion
                selectedNamespace={formState.credential.namespace}
```

jasonmadigan (changes requested):
Most of the last round is sorted: Secret label, `allowAlreadyExists: false`, the existing-route filter on `networking.istio.io/Hostname`, DR prefill from the ServiceEntry, YAML revalidation in all four steps, and the internal wizard's button labels are back. Thanks.

Three left. Two of them break things that work on `main`:

1. `showNamespaceField={false}` is now unconditional, which kills the internal wizard's default existing-route path and 4 tests in `mcp-wizard.spec.ts`.
2. `handleAddRule` seeding a match removes the `Add match` button, and 3 tests in `httproute-crud.spec.ts` still click it.
3. The external wizard still builds the registration from `formState.server.namespace`.

jasonmadigan:
:eyes:

src/components/httproute/HTTPRouteCreatePage.tsx
```diff
@@ -365,7 +367,7 @@ const HTTPRouteCreatePage: React.FC<HTTPRouteCreatePageProps> = ({ onFormChange
     setEditingRuleIndex(null);
     setCurrentRule({
       id: `rule-${Date.now().toString(36)}`,
-      matches: [],
+      matches: [{ id: 'match-1', pathType: 'PathPrefix', pathValue: '/', method: 'GET' }],
```
jasonmadigan: This seeds every new rule with `method: 'GET'`, and `matchUtils.ts` emits `method` whenever it is set. `main` had `matches: []`, so the serialisation branch never fired before, which makes this a behaviour change rather than just a new default.

`MCPSetupWizard` and `MCPRegistrationWizard` both embed this page, so routes created through those two become GET-only as well, and MCP JSON-RPC is POST (`src/utils/mcp/client.ts:308`).

Could we drop `method` from the seed and keep `pathType`/`pathValue`? That keeps the prefill improvement over main without pinning the method.

jasonmadigan (commented):
One thing worth catching before this goes in, otherwise this looks good to me.

jasonmadigan (approved):

jasonmadigan (approved):

Issue #795, opened 2026-08-27
MCP Setup Wizard: Gateway created in Step 1 is not selectable as HTTPRoute parentRef in Step 2


Anton-Fil:
---

Component

src/components/mcp/ — MCP Management setup wizard

Summary

In the MCP Setup Wizard, all resources (Gateway, HTTPRoute, ReferenceGrant, MCPGatewayExtension) are created only at the final Verify step (Step 4). Because of this, a Gateway that the user chooses to create in Step 1 does not yet exist in the cluster when they reach Step 2 (Route for Gateway). The embedded HTTPRoute create form populates its parentRef selector from a live cluster watch, so the just-defined Gateway is missing from the dropdown and cannot be selected as the route's parent — which is the whole point of the wizard flow.

Steps to Reproduce

1. Go to MCP Management with no existing MCPGatewayExtension → land on the Get started empty state.
2. Click MCP gateway setup wizard (/kuadrant/mcp/setup-wizard).
3. Step 1: choose Create a new Gateway, fill in a name (e.g. my-mcp-gw).
4. Step 2: choose Create a new HTTPRoute and open the parentRef selector.
5. The Gateway my-mcp-gw from Step 1 is not in the list of selectable parents.

Expected

The Gateway defined in Step 1 should be selectable as the HTTPRoute's parentRef in Step 2.

Actual

The Step 1 Gateway is absent from the parentRef selector because it is only created at Step 4 (Verify).

Root Cause

- MCPSetupWizard.tsx collects new resources into verifyItems (newGatewayResource, newRouteResource, etc.) and defers all k8sCreate calls to MCPVerifyStep.tsx (Step 4).
- The Step 2 HTTPRoute form is the embedded HTTPRouteCreatePage, whose parent selector (GatewaySelect / ParentReferencesSelect) reads Gateways via useK8sWatchResource (live cluster state) — see GatewaySelect.tsx:44.
- Since the Step 1 Gateway hasn't been persisted yet, the watch never returns it.

Proposed Solution

Rework the wizard so the Gateway is created during Step 1 (on "Next"), rather than deferred to the Verify step. Once persisted, the live watch in Step 2 will surface it as a selectable parentRef automatically.

Design points to resolve during implementation:
[...]

PR #811, opened 2026-08-31
Adding in the mcpgatewayexstension httproute management filed and tog…

R-Lawton:
## Summary

This PR updates MCP setup and management entry points.

### Changes

- Add optional HTTPRoute management to the MCP setup wizard. Automatic management remains enabled by default; disabling it exposes a step to choose or create an HTTPRoute.
- Keep the selected or newly created HTTPRoute connected to the MCPGatewayExtension configuration and verification flow.
- Add a populated MCP management getting-started panel with three compact entries:
  - Kuadrant documentation
  - MCPGatewayExtension setup wizard
  - MCPServerRegistration wizard
- Keep the existing empty-state setup wizard when no MCP resources exist.
- Make the new wizard actions compact link buttons so they are distinct from resource-level Create/Register actions.
- Improve E2E coverage for both gateway input variants and multiple MCP resources, and remove duplicate HTTPRoute switch rendering.

### Verification

Checks completed locally:

- `./node_modules/.bin/eslint src`
- `./node_modules/.bin/stylelint src/**/*.css --allow-empty-input`
- `./node_modules/.bin/tsc --noEmit`
- `./node_modules/.bin/jest --runInBand --roots src/components/mcp` — 94 tests passed
- `bash build/check-spec-map.sh`
- `./node_modules/.bin/playwright test --config=e2e/playwright.config.ts e2e/tests/mcp-setup-wizard.spec.ts --list`

For manual verification, start the local OpenShift Console environment, verify the existing empty-state setup wizard when no MCP resources exist, then verify the three getting-started entries after MCP resources are present. Open both wizard actions, and disable automatic HTTPRoute management to verify the HTTPRoute step appears and supports choosing or creating a route.

## Summary by CodeRabbit

* **New Features**
  * Reworked the MCP setup wizard into three steps, with optional HTTPRoute configuration.
  * Added an option to disable automatic HTTPRoute management.
  * Added guided setup choices for documentation, MCP extensions and server registrations.
  * Added support for cross-namespace gateways and required gateway references.
  * Added clearer validation messages for gateway, listener, secret and OAuth settings.

* **Bug Fixes**
  * Improved verification errors and access-denied feedback.
  * Improved HTTPRoute filtering and gateway association validation.

* **Style**
  * Refined MCP overview layouts and responsive behaviour on smaller screens.


Anton-Fil:
👀

Anton-Fil (commented):
Blocking — `src/components/mcp/MCPGatewayExtensionCreatePage.tsx:245`

`isMCPGatewayExtensionValid` now validates that the selected listener exists when `selectedGateway` is provided, but the standalone page calls it without that argument. This allows an `MCPGatewayExtension` to be created with a listener that does not exist on the selected Gateway. Could we pass `selectedGateway` here?

```tsx
validation={isMCPGatewayExtensionValid(formState, selectedGateway)}
```

Nit — `src/components/mcp/MCPGatewayExtensionCreatePage.tsx:234-242`

The standalone page does not pass `validationError` to `MCPExtensionFormFields`. When the session Secret, OAuth server, or listener is invalid, the save button is disabled but the user is not shown the reason. Could we calculate and pass the validation error?

```tsx
const validationError = getMCPGatewayExtensionValidationError(
  formState,
  selectedGateway,
);
```

```tsx
<MCPExtensionFormFields
  formState={formState}
  updateFormState={updateFormState}
  selectedGateway={selectedGateway}
  validationError={validationError}
/>
```

Anton-Fil (changes requested):

src/components/mcp/MCPExtensionFormFields.tsx
```diff
@@ -196,6 +201,13 @@ const MCPExtensionFormFields: React.FC<MCPExtensionFormFieldsProps> = ({
           placeholder={t('Enter extension name')}
           data-test="mcp-extension-name"
         />
+        {validationError?.field === 'extensionName' && formState.extensionName.trim() && (
```
Anton-Fil: Duplicate error message for the same field. This new validationError block renders alongside the pre-existing blur-driven message just below it (line 214, touched.extensionName && errors.extensionName). Typing an invalid name (e.g. UPPER) and blurring shows two red errors with different wording. Please render a single source of truth. Same pattern applies to targetGateway (line 319) and sectionName (line 381), which also have blur-driven errors twins.

src/components/mcp/MCPOverviewPage.tsx
```diff
@@ -486,9 +510,13 @@ const MCPOverviewPage: React.FC = () => {
     );
   }
 
-  const hasNoExtensions = extensionsLoaded && (!extensions || extensions.length === 0);
+  const hasNoMcpResources =
+    extensionsLoaded &&
+    serversLoaded &&
```
Anton-Fil: Empty-state regression. Gating hasNoMcpResources on serversLoaded means that if the MCPServerRegistration CRD is absent or the user lacks list permission, serversLoaded never becomes true, hasNoMcpResources stays permanently false, and the onboarding "Get started" empty state is never shown (previously it depended only on extensionsLoaded). Please treat an unloaded/errored servers watch as "no servers" for the empty-state check.

Anton-Fil (commented):
Small thinks to fix

src/utils/ParentReferencesSelect.tsx
```diff
@@ -415,24 +376,21 @@ const ParentReferencesSelect: React.FC<ParentReferencesSelectProps> = ({
...
                     {getSortedGateways().map((gateway) => {
                       const restriction = validateGateway(gateway);
-                      const gatewayKey = `${gateway.metadata.namespace}/${gateway.metadata.name}`;
 
                       return (
                         <FormSelectOption
-                          key={gatewayKey}
-                          value={gatewayKey}
+                          key={`${gateway.metadata.name}-${gateway.metadata.namespace}`}
+                          value={gateway.metadata.name}
```
Anton-Fil: Gateway options are resolved by name only, which reintroduces the #795 bug. Both the option `value` (line 393) and the `FormSelect` value (line 379) are `gateway.metadata.name`, and `updateParentReference` resolves the selection with `.find((gw) => gw.metadata.name === value)` (line 273). Since `metadata.name` is not unique across namespaces, two Gateways named e.g. `admin` in `team-a` and `team-b` produce two `<option>`s with the same `value="admin"`. Selecting the second one fires `onChange('admin')`, `.find` returns the first match, and the wrong `gatewayNamespace` is written into the parentRef — the HTTPRoute silently attaches to the wrong Gateway (reproducible in All Namespaces mode). This is exactly what #795 fixed with the composite `${namespace}/${name}` key and the dedicated `updateParentGateway` handler. Please restore composite `(namespace, name)` identification for the option/select value and resolution, keeping `requiredParentRef` supported.

src/components/mcp/MCPSetupWizard.tsx
```diff
@@ -397,150 +465,139 @@ const MCPSetupWizard: React.FC = () => {
...
+                    </div>
+                  </CardBody>
+                )}
+              </Card>
+            </WizardStep>
+          )}
+
+          {/* Verify configuration */}
+          <WizardStep
+            name={t(isRouteStepVisible ? '4. Verify configuration' : '3. Verify configuration')}
```
Anton-Fil, line 595: `t('4. Verify configuration')` / `t('3. Verify configuration')` are used here but neither key exists in `locales/en/plugin__kuadrant-console-plugin.json` — this PR removed `"4. Verify configuration"` and only `"Verify configuration"` remains. i18next will fall back to the raw key (looks fine in English), but the i18n consistency check in `test-frontend.sh` will fail and the strings won't be extracted for translation. Please run `yarn i18n` so both keys land in the locale (or use a single stable key with the step number as a param).

src/components/mcp/MCPOverviewPage.tsx
```diff
@@ -267,6 +277,20 @@ const MCPOverviewPage: React.FC = () => {
...
+  };
+
+  const httpRouteRBAC = {
+    list: userRBAC[`${getResourceNameFromKind('HTTPRoute')}-list`],
+    create: userRBAC[`${getResourceNameFromKind('HTTPRoute')}-create`],
+  };
+
+  const canOpenExtensionWizard = extensionRBAC.create && (gatewayRBAC.list || gatewayRBAC.create);
+  const canOpenRegistrationWizard =
+    serverRBAC.create && (httpRouteRBAC.list || httpRouteRBAC.create);
```
Anton-Fil: Blocking — `src/components/mcp/MCPOverviewPage.tsx:480`

The overview returns a full-page `Access Denied` whenever `!extensionRBAC.list`. A user who can list `MCPServerRegistration` but not `MCPGatewayExtension` is now locked out of the entire page — they can't see MCP servers or open the server registration wizard, even though the per-card access-denied states below already handle this gracefully. Since this PR reworks server-side RBAC here (`serverRBAC`, `canOpenRegistrationWizard`), the page-level gate should be relaxed to only deny when the user can list neither resource, keeping the per-card states intact:

```tsx
if (!extensionRBAC.list && !serverRBAC.list) {
  return ( /* Access Denied */ );
}
```

Anton-Fil (changes requested):
Two remaining blockers from re-review (the i18n Verify-step keys flagged earlier are still unaddressed as well).

src/components/mcp/MCPOverviewPage.tsx
R-Lawton: good catch

R-Lawton:
bad rebase removed fixes asked for and reintroduced work ill fix that thanks Anton

Anton-Fil:
Pr back "Done" and "Cancel" button that was removed in #809
image

R-Lawton:
Screenshot 2026-09-23 at 14 47 24
it has been removed im not sure i get what your asking @Anton-Fil ?

Anton-Fil:
> Screenshot 2026-09-23 at 14 47 24 it has been removed im not sure i get what your asking @Anton-Fil ?

yes, u fix problem

Anton-Fil (commented):
Fresh re-review of the HTTPRoute-management refactor. Application logic, i18n keys, imports and types check out. Two items below: a blocking one (stale e2e regression tests that will fail CI) and a non-blocking cleanup.

e2e/tests/mcp-setup-wizard.spec.ts
```diff
@@ -658,7 +671,20 @@ spec:
...
+          await sectionSelect.selectOption('mcp');
+        } else {
+          await page.locator('[data-test="mcp-section-name-input"]').fill('mcp');
+        }
+        await nextButton.click();
+
+        // HTTPRoute step: create a new route and select the draft Gateway as parentRef
         await expect(page.getByLabel('Create a new HTTPRoute')).toBeVisible({ timeout: 15_000 });
         await page.getByLabel('Create a new HTTPRoute').click();
         await expect(page.locator('#httproute-name')).toBeVisible({ timeout: 15_000 });
```
Anton-Fil: The two regression tests just below (they sit outside the diff, so I'm anchoring here) are stale against the new wizard flow and will fail. Since this spec file is directly edited, `build/suite-router.sh` runs it without `--grep @smoke`, i.e. with all tags — so these `@nightly` tests run in this PR's CI, not just on the nightly schedule.

- `Gateway draft state is preserved when navigating back to Step 1` (L715–750): L738 expects the `Choose or create an HTTPRoute` heading right after Step 1 → Next, but Step 2 is now `MCP Extension`. The Gateway draft itself is still preserved (`GatewayCreatePage` still receives `initialResource`), so only the intermediate assertion needs fixing.
- `HTTPRoute draft state (incl. parentRef) is preserved when navigating back to Step 2` (L752–804): L777 expects `Create a new HTTPRoute` on Step 2 (the route step is hidden by default now), and L796 clicks the removed `2. Route for Gateway` step. Its premise — restoring an HTTPRoute draft — is also gone: this PR removed `initialResource` / `hasHydratedFromInitial` / `skipInitialEmit` from `HTTPRouteCreatePage`. This one should be removed or rewritten to first disable automatic HTTPRoute management.

src/components/mcp/MCPExtensionStep.tsx
```diff
@@ -33,6 +36,11 @@ const MCPExtensionStep: React.FC<MCPExtensionStepProps> = ({
     () => buildMCPGatewayExtension(formState, selectedNamespace),
     [formState, selectedNamespace],
   );
+  const validationError = getMCPGatewayExtensionValidationError(formState, selectedGateway);
```
Anton-Fil: `getMCPGatewayExtensionValidationError(formState, selectedGateway)` is computed here and again in `MCPSetupWizard` (`extensionValidationError`). The wizard also no longer passes `onValidationChange` to this step, so `handleValidationChange` and the `onValidationChange` plumbing are effectively dead on the wizard path. Not a bug, but consider computing the validation error once and passing it down (or dropping the now-unused callback wiring) to avoid the duplicate compute.

Anton-Fil (approved):
LGTM

PR #849, opened 2026-09-09
fix(skill): handle Console runtime constraints

jasonmadigan:
## Summary

- investigate SDK peer-range patches before classifying shared-module bumps as impossible
- allow maintainer replacements only for proven metadata-only mismatches
- track runtime exclusions in Dependabot config and close blocked PRs after exclusions merge
- require validation against the oldest supported Console runtime before widening SDK peer ranges

## Verification

- `git diff --check upstream/main...HEAD`
- independent skill review found no remaining high or medium correctness issues


.claude/skills/dependabot-pr-maintenance/SKILL.md
```diff
@@ -25,8 +25,9 @@ Defer without acting when any of: label `do-not-merge/*` or `maintainers-only`;
...
+- `@openshift-console/dynamic-plugin-sdk` and `@openshift-console/dynamic-plugin-sdk-webpack` are pinned to the OCP line the plugin targets (the exact version lives in `package.json`), and the SDK is patched through yarn `patch:` (`.yarn/patches/README.md`). Never merge a Dependabot bump of `@openshift-console/*`: the hold is durable, so open the `dependabot.yml` exclusion PR described below and close the Dependabot PR after that exclusion merges.
+- The Console provides modules named by the SDK's `peerDependencies` at runtime as federation singletons. Read that manifest rather than relying on a cached package list. `ConsoleRemotePlugin` fails `yarn build` when the resolved version leaves an SDK peer range, but that error alone does not prove the bump is impossible. Inspect `docs/versioning.md`, the SDK section of `RELEASE.md`, `.yarn/patches/`, `.yarn/patches/README.md`, `src/__tests__/sharedModules.test.ts`, the bumped package's peer requirements, and its API and release changes.
```
Anton-Fil, line 29: nit: "Read that manifest" is less actionable than the concrete path this replaced. Since the audience is an LLM executing the skill, an explicit path lowers the chance it reads the wrong file or guesses. Dropping the module list is fine (it goes stale), but consider keeping the path, e.g.:

> Read the SDK's `peerDependencies` from `node_modules/@openshift-console/dynamic-plugin-sdk/package.json` rather than a cached package list.

_🤖 From Claude Code review (non-blocking)_

.claude/skills/dependabot-pr-maintenance/SKILL.md
```diff
@@ -58,7 +60,7 @@ Use this path for updates blocked by malformed generation, `resolutions` entries
...
 
-Durable exclusions go in `.github/dependabot.yml` `ignore` through a normal PR. A comment `@dependabot ignore DEPENDENCY major version` is for a proven, one-off incompatibility only; it closes the PR (a grouped PR is regenerated without that dependency later) and must appear in the report. Never create a replacement PR for a compatibility gate that explicitly forbids the target version.
+Runtime compatibility exclusions go in `.github/dependabot.yml` through a normal PR. Add a YAML comment with current runtime evidence and the SDK/OCP release that should remove the exclusion. Leave the blocked dependency PR open until the exclusion merges, then close it; an existing exclusion PR is its only valid temporary blocker. For an isolated bad package release unrelated to the target runtime, use the narrowest live command its PR body documents (`patch`, `minor` or `major`); name only the incompatible dependency in a grouped PR so Dependabot regenerates the group without it. Never create a replacement PR for a compatibility gate that explicitly forbids the target version.
```
Anton-Fil, line 63: nit: this dropped the explicit `ignore` YAML key (previously "`.github/dependabot.yml` `ignore`") and the exact `@dependabot ignore … version` command name. For an agent executing the skill, that concreteness matters — the current `dependabot.yml` has no `ignore` example to copy from, so it has to infer both the key and the command syntax. Consider keeping the key, e.g.:

> Runtime compatibility exclusions go in `.github/dependabot.yml` under `ignore` through a normal PR.

_🤖 From Claude Code review (non-blocking)_

PR #859, opened 2026-09-16
Namespace Flickering Bug & Landing

### LINKED ISSUE #847: NamespaceBar flickering after creating a new namespace
## Description

When selecting a new namespace in the **Kuadrant Overview** page via the `NamespaceBar`, the **namespace selection flickers.** The page briefly shows the new namespace, then flickers the previous namespace, jumps back to the previous resources/namespace. 

### Steps to Reproduce

1. Navigate to Kuadrant → Overview
2. Note the currently selected namespace in the NamespaceBar
3. Click on the NamespaceBar dropdown and select **"Create namespace"**
4. Enter a name for the new namespace and create it
5. Observe the page content during the transition after the new namespace is created

### Expected behavior

- After creating the new namespace, the Overview page should immediately switch to show the newly created namespace
- The transition should be smooth and direct
- The new namespace should appear in the NamespaceBar and remain stable
- Gateway counts, policy lists, and HTTPRoutes for the new (empty) namespace should display once

### Actual behavior

- The new namespace's resources briefly appear
- The page flickers back to the previous namespace for ~1 second
- The page finally settles on the previous namespace

### Screenshots

https://github.com/user-attachments/assets/329de62f-7e51-467a-9100-d8e599ac500b

**Affected Files:**
- src/components/KuadrantOverviewPage.tsx
- src/hooks/useKuadrantNamespaceChange.ts

**Related Issues:**
#824 (namespace switching crashes) — Different issue but both involve namespace transition lifecycle

pascazzioIsAbyayalan:
## Description

Namespace selection flickered when creating a new namespace via `NamespaceBar:` _**old-ns → new-ns → old-ns**_, landing back on the old namespace instead of the newly created one. 

**Root cause:** Three competing React effects managing namespace state, plus incorrect sync direction in `useKuadrantNamespaceChange` hook and component calls `setActiveNamespace` directly when creating a namespace, expecting the app to navigate to match.

Fixed by reversing sync direction (activeNamespace → URL navigation) and removing redundant effects

Fixes #847 

## Type of change

- [ ] `[fix]` Bug fix (non-breaking change that fixes an issue)

## Changes made

- Reversed sync direction: **Changed from URL → activeNamespace** (reset) to **activeNamespace → URL** (navigate)
- When `activeNamespace` changes, hook now navigates URL to match instead of resetting `activeNamespace` back to URL
- Respects `NamespaceBar` component's design intent: `setActiveNamespace` triggers navigation
- Removed unused setter: `setActiveNamespace `no longer needed after reversing sync direction
- Removed redundant sync effect: Deleted competing effect that called `handleNamespaceChange(ns)` on URL changes
- Fixed redirect effect dependencies: Removed activeNamespace from deps to prevent re-runs during namespace transitions
- Hook (`useKuadrantNamespaceChange`) is now the single source of truth for namespace state management

## Test plan

- [ ] `yarn lint` passes (no changes after running)
- [ ] `yarn build` passes
- [ ] `yarn i18n` passes (no changes after running — commit updated locale files if needed)
- [ ] Tested manually in OpenShift Console
- [ ] Tested in both light and dark themes
- [ ] Tested in all-namespaces and single namespace mode
- [ ] Create new namespace via NamespaceBar → navigates to new namespace without flicker
- [ ] Switch between existing namespaces → smooth transition
- [ ] Switch to all-namespaces view → navigates correctly

## Screen record

https://github.com/user-attachments/assets/a000ddff-aee5-4e1c-87af-006c2baca888

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved navigation between Kuadrant namespaces so the URL remains aligned with the selected namespace.
  - Selecting a specific namespace now routes to its corresponding namespace URL.
  - Selecting the all-namespaces view now routes to the dedicated all-namespaces page.
  - Prevented unnecessary namespace updates and redirect loops during URL changes.
  - Improved redirect handling after permission checks, preventing navigation after cancellation or completion.
  - Ensured namespace synchronisation works correctly with direct and in-app navigation.

- **Documentation**
  - Clarified that namespace transition and permission-redirect coverage is provided by end-to-end tests.


src/hooks/useKuadrantNamespaceChange.ts
```diff
@@ -28,20 +28,25 @@ import { useActiveNamespace } from '@openshift-console/dynamic-plugin-sdk';
  */
 export const useKuadrantNamespaceChange = (basePath: string) => {
   const { ns } = useParams<{ ns: string }>();
-  const [activeNamespace, setActiveNamespace] = useActiveNamespace();
+  const [activeNamespace] = useActiveNamespace();
```
jasonmadigan: This was the only place activeNamespace got written from the URL. The console doesn't derive it from `/kuadrant/...` paths (`getNamespace` only parses `/<x>/ns/<ns>`), so deep links and in-app navigation to `/ns/X` now bounce to whatever the console last had. That's the smoke failure: every test lands on `/all-namespaces`. MyAPIKeysPage and MCPOverviewPage still have the page effect removed here, so there the URL sticks but activeNamespace stays stale (Request API Key disabled, wizard routes empty). Suggestion: keep URL to state keyed on `[ns]`, add state to URL keyed on `[activeNamespace]`, each behind a previous-value ref so it only reacts to its own change, subpath preserved. Worth a spec for the deep-link and create-namespace paths so CI guards this.

src/hooks/useKuadrantNamespaceChange.ts
```diff
@@ -28,20 +28,25 @@ import { useActiveNamespace } from '@openshift-console/dynamic-plugin-sdk';
...
   React.useEffect(() => {
-    if (ns && ns !== activeNamespace) {
-      setActiveNamespace(ns);
-    } else if (!ns && activeNamespace !== allNamespacesSubPath) {
-      setActiveNamespace(allNamespacesSubPath);
+    // If activeNamespace changed but URL doesn't match, navigate to match activeNamespace
+    if (activeNamespace && activeNamespace !== allNamespacesSubPath && activeNamespace !== ns) {
+      const targetUrl = `/kuadrant${basePath}/ns/${activeNamespace}`;
+      navigate(targetUrl, { replace: true });
+    } else if (activeNamespace === allNamespacesSubPath && ns !== 'all-namespaces') {
```
jasonmadigan: `ns` is undefined on the all-namespaces routes (no `:ns` in `console-extensions.json`), so this is always true, and `navigate` changes identity on every pathname change under BrowserRouter. With All Namespaces selected, each tab click under `/all-namespaces/<tab>` gets replaced with the bare route and tab deep links lose the tab. Might be worth comparing against the full pathname (subpath included) rather than `ns`.

src/components/KuadrantOverviewPage.tsx
```diff
@@ -345,7 +345,7 @@ const KuadrantOverviewPage: React.FC = () => {
     };
 
     performRedirect();
-  }, [location.pathname, activeNamespace, navigate]);
+  }, [location.pathname, navigate]);
```
jasonmadigan, line 385: With All Namespaces selected and no cluster-wide gateways list access, this redirects to `/ns/default`, the hook replaces back to `/all-namespaces`, the pathname change re-runs this, and so on (about 1000 navigations in 8s locally; main settles in one hop). `rbac.spec.ts:43` only passes because `toHaveURL` polls. MyAPIKeysPage has the same pair. Also `activeNamespace` is still read after `await checkAccess` (334, 341), so dropping the dep leaves a stale closure rather than preventing re-runs; the hook changes pathname on every transition anyway. A cancellation flag in the cleanup would cover it.

jasonmadigan (changes requested):
Some comments inline.

jasonmadigan:
the e2e-rbac failure is this change, not flake: a fresh CI session starts at All Namespaces, every spec navigates to `/kuadrant/<page>/ns/kuadrant-test`, and the hook bounces it to `/all-namespaces` (traces in https://github.com/Kuadrant/kuadrant-console-plugin/actions/runs/35129088650).

pascazzioIsAbyayalan:
@jasonmadigan  I addressed yours and CodeRabbit comments: 

- Implemented bidirectional sync with loop guards
- Compared full pathname instead of just ns
- Redirected Loop
   - Cancellation flag for cleanup
   - One-time RBAC check
   - Prevent hook from fighting redirect
- Added comprehensive documentation in KuadrantOverviewPage.test.tsx

src/hooks/useKuadrantNamespaceChange.ts
```diff
@@ -33,16 +33,75 @@ export const useKuadrantNamespaceChange = (basePath: string) => {
...
+  // Initialize with sentinels so first render detects changes
+  const prevNsRef = React.useRef<string | undefined>(undefined);
+  const prevActiveNamespaceRef = React.useRef<string | undefined>(undefined);
+
+  // Bidirectional sync 1: URL → activeNamespace
+  // When URL changes (e.g., deep links, in-app navigation), update activeNamespace
   React.useEffect(() => {
-    if (ns && ns !== activeNamespace) {
+    // Only react if ns changed (not if activeNamespace changed)
+    if (ns === prevNsRef.current) return;
```
jasonmadigan, line 45: `prevNsRef` starts as `undefined`, and `ns` is also `undefined` on the `/all-namespaces` routes (there is no `:ns` there), so on those routes this returns on mount and `#ALL_NS#` is never set. With a namespace selected, a deep link to `/kuadrant/policies/all-namespaces` then gets bounced to `/ns/<selected>` by the effect below. The version in the next comment avoids this by deriving `urlNamespace = ns || allNamespacesSubPath`.

src/hooks/useKuadrantNamespaceChange.ts
```diff
@@ -33,16 +33,75 @@ export const useKuadrantNamespaceChange = (basePath: string) => {
...
+    } else {
+      expectedUrl = `/kuadrant${basePath}/all-namespaces${subPath}`;
+    }
+
+    // Only navigate if current pathname doesn't match expected URL
+    // Special case: if activeNamespace is #ALL_NS# but we're on a namespace path (e.g., /ns/default),
+    // don't navigate back to /all-namespaces - let the URL→state effect sync activeNamespace instead
+    // This prevents redirect loops when RBAC redirects user to a namespace
+    const isOnNamespacePath = location.pathname.includes(`${basePath}/ns/`);
+    if (activeNamespace === allNamespacesSubPath && isOnNamespacePath) {
```
jasonmadigan, line 95: This branch is where the smoke run goes wrong. Shortly after we set `activeNamespace`, the console resets it to `#ALL_NS#`: our `setActiveNamespace` updates the console's last-namespace preference, its detect-namespace effect re-runs (console `getNamespace` only understands `/<x>/ns/<ns>`, so on `/kuadrant/...` it sees no URL namespace), `checkNamespaceExists` GETs the Namespace, and the e2e personas cannot GET namespaces (test-admin included), so it falls back to `#ALL_NS#`. This branch then bails, and the URL to state effect above never re-runs because `ns` did not change. Net result: URL `/ns/kuadrant-test`, state `#ALL_NS#`. The policies page checks RBAC against `default` (missing tabs and Create button for test-viewer and test-devops), and My API Keys renders the `Namespace` column (the strict-mode violation in data-view-regressions). `main` only survives because its effect depends on `activeNamespace` and re-asserts the URL namespace on every change.

Suggest collapsing both effects into one where the URL wins and the two external cases (console fallback, namespace created from the NamespaceBar) are handled explicitly. Drop both effects and the two `prev*Ref`s, and place this after `getCurrentSubPath` so it can reuse it:

```ts
  // namespace the url expresses: /ns/:ns, otherwise all namespaces
  const urlNamespace = ns || allNamespacesSubPath;

  // last namespace url and console state agreed on
  const syncedRef = React.useRef<string | undefined>(undefined);
  // url namespace already re-asserted once against a console reset
  const reassertedRef = React.useRef<string | undefined>(undefined);

  React.useEffect(() => {
    if (urlNamespace !== syncedRef.current) {
      // url changed: deep link, tab, rbac redirect, NamespaceBar select. url wins
      syncedRef.current = urlNamespace;
      reassertedRef.current = undefined;
      if (activeNamespace !== urlNamespace) {
        setActiveNamespace(urlNamespace);
      }
      return;
    }

    if (activeNamespace === urlNamespace) return;

    if (activeNamespace === allNamespacesSubPath) {
      // console fell back to all namespaces (namespace get failed). re-assert url once,
      // then accept the console's decision rather than loop against it
      if (reassertedRef.current !== urlNamespace) {
        reassertedRef.current = urlNamespace;
        setActiveNamespace(urlNamespace);
      }
      return;
    }

    // namespace created or selected outside handleNamespaceChange: follow it
    syncedRef.current = activeNamespace;
    navigate(`/kuadrant${basePath}/ns/${activeNamespace}${getCurrentSubPath()}`, {
      replace: true,
    });
  }, [
    urlNamespace,
    activeNamespace,
    setActiveNamespace,
    navigate,
    basePath,
    allNamespacesSubPath,
    getCurrentSubPath,
  ]);
```

URL change (deep link, tab, RBAC redirect, NamespaceBar select) sets state. A state change to a concrete namespace with an unchanged URL is the create flow and navigates with the tab kept (#847). State falling back to `#ALL_NS#` on a `/ns/` URL is re-asserted once, then accepted, so we never loop against the console (the dev perspective NamespaceBar runs the same GET check and would otherwise ping-pong). I ran this through a small jest harness covering deep link plus console reset, `/all-namespaces` with a namespace selected, create namespace, RBAC redirect, tab click and NamespaceBar select: all pass, `main` fails only the create case, this branch fails deep link, `/all-namespaces` and RBAC redirect.

src/components/KuadrantOverviewPage.tsx
```diff
@@ -318,16 +318,32 @@ const KuadrantOverviewPage: React.FC = () => {
   const watchNamespace = ns || activeNamespace;
 
   // Smart default redirect: check cluster-wide permissions and redirect namespace-scoped users
+  // Track if we've already performed the RBAC check to prevent redirect loops
+  const rbacCheckPerformedRef = React.useRef(false);
```
jasonmadigan, line 322: Once the hook stops navigating on `#ALL_NS#`, the redirect loop this guards against is gone, so `rbacCheckPerformedRef` and `checkPending` can go; `cancelled` is worth keeping. Rather than dropping `activeNamespace` from the deps (lines 350 and 360 then read a stale closure), read it through a ref:

```tsx
  // Smart default redirect: check cluster-wide permissions and redirect namespace-scoped users
  const activeNamespaceRef = React.useRef(activeNamespace);
  activeNamespaceRef.current = activeNamespace;

  React.useEffect(() => {
    if (location.pathname !== '/kuadrant/overview/all-namespaces') return;
    let cancelled = false;
    const redirect = () => {
      const current = activeNamespaceRef.current;
      const target = current && current !== '#ALL_NS#' ? current : 'default';
      navigate(`/kuadrant/overview/ns/${target}`, { replace: true });
    };
    checkAccess({ group: 'gateway.networking.k8s.io', resource: 'gateways', verb: 'list' })
      .then((result) => {
        if (!cancelled && !result.status?.allowed) redirect();
      })
      .catch(() => {
        if (!cancelled) redirect();
      });
    return () => {
      cancelled = true;
    };
  }, [location.pathname, navigate]);
```

src/components/KuadrantOverviewPage.test.tsx
```diff
@@ -66,3 +66,18 @@ describe('ErrorCodeLabel', () => {
     expect(screen.queryByText('Code: 403')).not.toBeInTheDocument();
   });
 });
+
+/*
```
jasonmadigan, line 70: The e2e suite did not catch this and needs a cluster. A hook-level test catches it in seconds: mock `react-router` and `useActiveNamespace` as an external store, then simulate the console reset. I have a harness for the cases listed above; happy to share it here or push it to this branch, whichever you prefer.

jasonmadigan (changes requested):
Couple of test failures, suggested fixes inline below. thanks @pascazzioIsAbyayalan

src/components/KuadrantOverviewPage.test.tsx
pascazzioIsAbyayalan: You can share it here, thank you so much. I am also checking the rest of your review now to apply the changes and push

jasonmadigan (changes requested):
Rechecked `26cfefe`. The latest commit memoizes the Policies tabs and context, but the namespace hook is unchanged. These two issues from the earlier review still reproduce:

- [All-namespaces links](https://github.com/Kuadrant/kuadrant-console-plugin/pull/859#discussion_r4072429196): with a namespace already selected, opening `/kuadrant/policies/all-namespaces/auth` redirects to `/ns/<previous-namespace>/auth`. Both `ns` and `prevNsRef` start as `undefined`, so the URL-to-state effect skips initialization.
- [Console namespace resets](https://github.com/Kuadrant/kuadrant-console-plugin/pull/859#discussion_r4072429211): after opening `/ns/kuadrant-test`, a console reset to `#ALL_NS#` leaves that state in place while the URL still says `kuadrant-test`. The early return assumes the other effect will restore it, but that effect skips because `ns` hasn't changed.

Both checks fail in focused automated tests against this commit; creating a new namespace and retaining the tab passes. Could we fix these two synchronization cases before merging?

jasonmadigan (approved):

emmaaroche (approved):

